PR #85, opened 2018-12-08
Add support to parse strings and multiple times

alxhotel:
This PR adds the following features:

- Currently the only way to parse cQASM code is by storing it in a `file`. This PR adds support to also parse `strings` of cQASM.

- If the user of `libqasm` wants to parse multiple files in one execution, the parser needs to be reset. This PR cleans the state of the previous parsing before the next one.


kel85uk:
Closing as it seems no further work was done to correct the changes.

PR #91, opened 2019-05-29
Add support for python interface

### LINKED ISSUE #89: Python usage of libqasm
Hello!

I was wondering how I could use libqasm to get the a QASM circuit in a python data structure which I could access and operate on.

I tried:

`
from libQasm import libQasm as lq

a = lq("./test/bell.qasm") #while on the qasm_flex_bison folder
b = a.getQasmRepresentation()
`

So far so good.
But they I try to access it's methods, like numQubits, etc I can't:
`b.numQubits()` doesn't work, for example... The only attributes of 'b' appear to be:
`b.acquire(  b.append(   b.disown(   b.next(     b.own(`

Any ideas?

alxhotel:
Fixes #89 

This PR adds support to use the rest of methods of `QasmRepresentation` from Python.

It also has a small styling refactor.


imranashraf:
@alxhotel right now there are no tests which check the interface. I will suggest

1. port the existing tests such that they utilize this python interface. this has advantages as they will not require a lot of re-compilation of c test files
2. secondly, some more tests should be added to check the extended interface
3. instead of using doctest for testing, we can use pytest

alxhotel:
Thanks for the review, @imranashraf. I agree :)

I think I can add those three points to the PR this week.

imranashraf:
that would be great @alxhotel . thanks

alxhotel:
Closing in favor of #96

PR #96, opened 2019-08-27
Make libqasm a reentrant parser with Python interface

oschusler:
Replaces pull request #91 

* Add python unittests next to the original doctests
* Adds additional tests for the extended interface to both doctest and python unittests

**Note**: We would like to keep the original doctests, since these test the c++ interfaces which are being used by other projects, like qx-simulator.


oschusler:
@imranashraf, is it correct that libqasm no longer tests on Jenkins, because it used to do before, right?

imranashraf:
@oschusler  No that is not correct.

imranashraf:
@oschusler the tests you have added still do not test the python interface being exposed to the user of libqasm, which was the main purpose of https://github.com/QE-Lab/libqasm/issues/89.

oschusler:
> @oschusler the tests you have added still do not test the python interface being exposed to the user of libqasm, which was the main purpose of #89.

Sorry about that, I didn't understand what you meant with the python interface. If I'm correct, we have added unittests for the entire interface. Could you please re-review and merge the PR?

PR #97, opened 2019-09-04
[DEM-1090] Windows Build

ammar92:


imranashraf:
@ammar92 I would suggest to use windows environment setup as is being used for openql. This is how people in Physics lab are using. This way we will not have to maintain multiple different setups

PR #98, opened 2019-11-18
[DEM-1188] init linenr and added tests for qasm as a string

QFer:
* yylineno and yycolumn are now initialised when using parse_string.
* added python-testcases in test_cqasm using parse_string
* adjusted assertRaisesRegex so that the Line: n part is also tested (escaping the or-symbol)


QFer:
When we put some white-spaces in front of the qubits definition as in the following program:

"""version 1.0
  qubits 1
X90 q[0]
Y90 q[0]
measure_all"""

we got "syntax error, unexpected WS, expecting COMMENT or QUBITS | Line: 538976268”

Because yylineno was not initialised the errormessage references a 'random' linenr.

PR #99, opened 2020-01-31
WIP: Rewrite setup to support all OS' and move to root

oschusler:
In order to be able to install libQasm properly as a python package, I have rewritten the `setup.py` file to install all files properly, and tried to make it as OS agnostic as possible. Other than that, I moved the `setup.py` file to the root, so that it can be installed from `setup.py` in different libraries. Finally, I have renamed the `qasm_flex_bison` to `src` since that is the normal designation for this type of folders.

@imranashraf Could you please review this PR, and comment on/approve the changes?


imranashraf:
@oschusler looks good. tested on both osx and nix. there is one small change thought which i will perge soon. thanks

PR #101, opened 2020-02-17
[DEM-NNN] Few fixes for build problems

QFer:
* updated README.md with correct path
* corrected README.md: unit test command and spelling error
* bugfix in libQasm.hpp parse_file file pointer was not checked
* include guards added to libQasm.hpp
* setup.py "make all" -> "make" to make all
* Changes in CMakeLists.txt
  * consistent uppercase commands
  * swig module for python now uses the static library
  * for successful running the tests, the dynamic library had to be copied to the test directory
  * some changes to prevent compilation warnings and makefile generation warnings
* Python import changes in unit tests
* fixed libQasm can be imported again in python


imranashraf:
works fine on linux but fails on osx. after building the module, clean the the src directory to remove build files and then test by importing libQasm.

QFer:
Hi Imran, 
I now link the static library for the python environment also for Darwin. Otherwise when using python, the dynamic library wasn't found, even when it is in the same directory as libQasm.so (just as I saw on Linux environment). Instead of setting a run time search path to the python site packages directory it is easier to link the static library also for darwin environment.

Is this solution oke with you?

src/library/CMakeLists.txt
```diff
@@ -75,17 +88,12 @@ FIND_PACKAGE(PythonLibs REQUIRED)
...
-
-if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-    SWIG_LINK_LIBRARIES(libQasm lexgram)
-    set_target_properties(_libQasm PROPERTIES LINK_FLAGS "-undefined dynamic_lookup")
-else ()
-    SWIG_LINK_LIBRARIES(libQasm ${PYTHON_LIBRARIES} lexgram)
-endif()
+ENDIF()
+
+SWIG_LINK_LIBRARIES(libQasm ${PYTHON_LIBRARIES} lexgramstatic)
```
imranashraf: linking python libraries is not required for osx, in fact segfaults when libQasm module is imported.

secondly, osx also requires link flag: -undefined dynamic_lookup. therefore, this line should be changed to:

```
IF (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
    SWIG_LINK_LIBRARIES(libQasm lexgramstatic)
    SET_TARGET_PROPERTIES(_libQasm PROPERTIES LINK_FLAGS "-undefined dynamic_lookup")
ELSE ()
    SWIG_LINK_LIBRARIES(libQasm ${PYTHON_LIBRARIES} lexgramstatic)
ENDIF()
```

QFer:
@imranashraf I changed the line to what you proposed. Tested it on an OSX machine and it succeeded. My previous solution also succeeded on that laptop, why you got segfault when importing libQasm I don't know.

imranashraf:
@QFer indeed it works.

regarding issue on my machine, i read somewhere that linking python libraries on osx was not recommended and also this second fix was discussed. I am not sure, may be it is because i am working with python under conda.

PR #104, opened 2020-05-28
Added skip n as copy of wait n, but meaning to not issue bundles in n…

jvansomeren:
Added skip n with n number of cycles, in syntax similar to wait n.
In the bundled representation of qasm, each line/bundle specifies
the instructions/gates that should start in one cycle.
Subsequent lines/bundles represent instructions/gates of subsequent cycles.
Instead of a bundle, the line can contain skip n, with n a number of cycles >= 1.
Meaning of skip n is to insert n cycles in which no instructions/bundles are started.
It is short for inserting n empty lines.

It was added because the current wait n is interpreted by the OpenQL compiler differently:
it means (standalone on a line) that execution should be scheduled such that all gates that
are still executing, complete before 'executing' the wait n; this is the barrier part of wait.
And that subsequently, there will be n cycles in which no gate is executing (so also not
in parallel).
And after that, the gates can be started that follow the wait n.
So wait n is a scheduling directive with a barrier semantics.

In an accompanying update, the OpenQL compiler will print skip n between bundles
instead of wait n when a skip n is intended.
And the cqasm reader will accept a skip n from libqasm and in an initial implementation
do nothing with it.
A later implementation of cqasm reader may use skip to support in setting the cycle
attributes of the created gates.

### DIFF
--- a/src/library/grammar.y
+++ b/src/library/grammar.y
@@ -52,7 +52,7 @@
 %token <sval> SINGLE_QUBIT_GATES TWO_QUBIT_GATES CR CRK TOFFOLI
 %token <sval> CDASH NOT_TOKEN
 %token <sval> MAPKEY PREP MEASURE MEASUREPARITY MEASUREALL
-%token <sval> WAIT DISPLAY RESET_AVERAGING LOAD_STATE QUOTED_STRING
+%token <sval> SKIP WAIT DISPLAY RESET_AVERAGING LOAD_STATE QUOTED_STRING
 %token <sval> ERROR_MODEL_KEY ERROR_MODEL
 %token QBITHEAD BITHEAD
 
@@ -419,8 +419,8 @@ parallelizable-ops : all-valid-operations
     ;
 
 //# Special operations
-%type <oval> special-operations display-operation wait-operation reset-averaging-operation load-state-operation;
-special-operations : display-operation | wait-operation | reset-averaging-operation | load-state-operation
+%type <oval> special-operations display-operation skip-operation wait-operation reset-averaging-operation load-state-operation;
+special-operations : display-operation | skip-operation | wait-operation | reset-averaging-operation | load-state-operation
     ;
 display-operation : DISPLAY NEWLINE
                     {
@@ -434,6 +434,11 @@ display-operation : DISPLAY NEWLINE
                     {
                         $$ = new compiler::Operation( std::string($1), *($3) );
                     }
+    ;
+skip-operation : SKIP WS INTEGER
+                 {
+                     $$ = new compiler::Operation( std::string($1,4), ($3) );
+                 }
     ;
 wait-operation : WAIT WS INTEGER
                  {
--- a/src/library/lex.l
+++ b/src/library/lex.l
@@ -52,6 +52,7 @@ comma_sep          {whitespace}?,{whitespace}?
 parallel_sep       {whitespace}?\|{whitespace}?
 qubits             (?i:qubits)
 mapkey             (?i:map)
+skip               (?i:skip)
 wait               (?i:wait)
 resetavg           (?i:reset-averaging)
 display            ((?i:display)|(?i:display_binary))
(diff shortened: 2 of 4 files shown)


razvnane:
> Meaning of skip n is to insert n cycles in which no instructions/bundles are started.
> It is short for inserting n empty lines.

What would happen in the following situation:
line1: Instruction1-3cycles ; Instruction2-5 cycles
line2: Skip 2
line3: Instruction3-2cycles

Would this be valid cqasm? 
 
> It was added because the current wait n is interpreted by the OpenQL compiler differently:

Could  we update the OpenQL compiler instead to behave as intended? 

> So wait n is a scheduling directive with a barrier semantics.

What can't we do using the existing wait syntax?

jvansomeren:
Op 3 jun. 2020, om 08:24 heeft razvnane <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:



Meaning of skip n is to insert n cycles in which no instructions/bundles are started.
It is short for inserting n empty lines.

What would happen in the following situation:
line1: Instruction1-3cycles ; Instruction2-5 cycles
line2: Skip 2
line3: Instruction3-2cycles

Would this be valid cqasm?

Yes, it would be valid cqasm; there are hardly any constraints on cqasm.
But what is missing is a skip 1 at the end to let the processor come to rest; but that could be assumed, since there is no ’stop’.
The point is usually whether it is valid eqasm, i.e. acceptable for the processor.

The effect of the above is:

cycle 1: instruction1 starts with 1st cycle; instruction2 starts with 1st cycle
cycle 2: no new instruction starts, instruction1 does 2nd cycle, instruction2 does 2nd cycle
cycle 3:  no new instruction starts, instruction1 does 3rd (final) cycle, instruction2 does 3rd cycle
cycle 4: instruction3 starts with 1st cycle; instruction2 does 4th cycle
cycle 5: instruction3 does 2nd (final) cycle; instruction2 does 5th (final) cycle
cycle 6: no instruction executes
cycle 7: no instruction executes
… … etc. etc.

It was added because the current wait n is interpreted by the OpenQL compiler differently:

Could we update the OpenQL compiler instead to behave as intended?

The OpenQL compiler knows both:
- it writes bundles, formerly with wait, now under an option with skip
- it reads bundles, formerly only with wait which are interpreted as wait gates with barrier semantics, now also with skip
So, I don’t understand your question.
The wait that has barrier semantics is actually used now in experiments (see discussions) so it cannot be replaced.
So when we would keep writing bundles with wait in it with the skip meaning, they keep being wrongly interpreted.
The only way out that preserves the wait with barrier semantics on reading is to add a skip.

So wait n is a scheduling directive with a barrier semantics.

What can't we do using the existing wait syntax?

You cannot drink beer with it. :-)
Current use is ’wait n’ with the described barrier meaning on reading.
And current use is ’wait n’ with the described skip meaning on writing.
We have a bug currently that the wait with barrier meaning is written as ’wait n’ as well.
So there is no distinction syntactically between the two,

jvansomeren:
The current state of dealing with wait to my knowledge is
that a wait gate is not preserved when writing (it is only visible in the cycle assignment) and that a wait n is written in the skip meaning.
So when scheduled IR is written and there is a wait, it means skip.
When non-scheduled IR is written and there is a wait, it means the wait gate.

On reading, any wait is interpreted as a wait gate.

So the IR is ill-defined when it is scheduled code.
It is only consistently written/read for non-scheduled code.

Until now, no-one cares about correct cycle assignment in reading scheduled code.
QX sees the parallel instructions in bundles but ignores there durations and cycle consequences.
And quantumsim upto recently only saw the OpenQL generated python script, not the cqasm with cycles.
CCLight assembly code is written by the OpenQL compiler with valid cycle information.

Op 3 jun. 2020, om 08:24 heeft razvnane <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:



Meaning of skip n is to insert n cycles in which no instructions/bundles are started.
It is short for inserting n empty lines.

What would happen in the following situation:
line1: Instruction1-3cycles ; Instruction2-5 cycles
line2: Skip 2
line3: Instruction3-2cycles

Would this be valid cqasm?

It was added because the current wait n is interpreted by the OpenQL compiler differently:

Could we update the OpenQL compiler instead to behave as intended?

So wait n is a scheduling directive with a barrier semantics.

What can't we do using the existing wait syntax?

razvnane:
Please reply on github to prevent automatic embedding of footers from the email into the message, making this unnecessarily long. I also cannot easily quote your message now.

Issue #94, opened 2019-07-21
clearing state across multiple runs


imranashraf:
compiler::SubCircuits subcircuits_object holds the state which is declared globally in grammar.y. This results in the stateful behavior which causes problem when library is used to load cqasm file multiple times.

alxhotel:
Yep. I also had this issue in the past. I see that you have already made a fix for it in #95.

What I ended up doing is an implementation for a reentrant scanner, as suggested by @kel85uk and @quantumkoen. And it seems to work.

If that is of any use to the repository, I would like to make a PR. But the implementation depends on #91. So #91 would have to be merged before.

imranashraf:
@alxhotel #91 seems to be interesting. currently checking it out. will merge is today in the afternoon.

jvanstraten:
I'm closing this, as I don't think it's a problem anymore since #105.

Issue #56, opened 2018-07-23
B as alias is not parsed


quantumkoen:
This code from the qx simulator test suite is not parsed:

```
version 1.0 

# file   : full_adder.qc
# author : Nader Khammassi
# brief  : quantum full adder (carry-save addition)

# define 4 qubits
qubits 4

# rename qubits
map q[0],A 
map q[1],B
map q[2],C
map q[3],D


# init inputs to some values
.init
  x A 
  x C 

  display_binary  #  initial input value


# perform addition
.add

  toffoli B,C,D
  cnot    B,C 
  toffoli A,C,D
  cnot    A,C 

  display_binary #  result

.reverse_add

  cnot    A,C 
  toffoli A,C,D
  cnot    B,C 
  toffoli B,C,D

  display_binary #  result

# error_model depolarizing_channel, 0.1 
```

Reports:

```
[+] loading circuit from 'tests/circuits/full_adder.qc' ...
Error while parsing file tests/circuits/full_adder.qc: 
syntax error, unexpected BITHEAD, expecting NAME
```

kel85uk:
Closing due to no response, and the way I interpret it is that b and q should be a special keyword for cQASM the same way as special registers (e.g. ax) are reserved for classical assembly.

quantumkoen:
Given that it is trivial to distinguish between the use of b as a signifier for adressing bits vs using it as an alias, and given that the cqasm 1.0 syntax nowhere says b cannot be used as an alias, I think it should be supported by the parser.

kel85uk:
True, reopening this.

jvanstraten:
This has been supported since #105, though due to a bug I stumbled on and fixed in #115 I think that particular mapping has only worked since then. Note that you wouldn't be able to refer to the measurement bits associated with the qubits once you redefine `b` though (unless you make another remap `b` first).

PR #105, opened 2020-06-26
Implement runtime-configurable instruction set for OpenQL, almost entirely rewriting the library

jvanstraten:
This is a major overhaul of the guts of libqasm.

Originally I was tasked to work on a feature request by the OpenQL compiler team, requesting that the set of instruction names could be set at runtime based on the OpenQL platform JSON file, or have it be possible to disable checking entirely. At the same time, I was given the requirement to be fully backward-compatible and have this be transparent to the end user.

In case the reader is unaware, the current bison grammar hardcodes the instruction names as keywords and does "typechecking" of the parameter lists grammatically, making it impossible to change the instruction set programmatically at runtime. It *might* have been possible in theory to replace the instruction keywords with identifiers in the current grammar (not 100% on this, might introduce a huge pile of grammatical ambiguity), but doing *just* that would break backward compatibility in the sense that any instruction would now be usable with any set of supported parameters. That could of course be moved to qasm_semantic.hpp and be disableable with a flag, but boy would that be ugly, and it wouldn't be reusable at all, because the instructions would still be hardcoded.

Combined with the fact that the grammar has always been a huge mess to begin with (whitespace and comments are not supposed to be tokens, for one thing!), the API is not much better (you're supposed to call delete for every new, and it's stateful for no real reason), and I already started working on a new version of libqasm one and a half year ago for my ultimately rejected cQASM 2.0 specification, I instead decided to just start over, and do it right this time.

Everything in `src/cqasm` is new, and can in fact be compiled and tested independently of everything else in the repository. Right now it's C++ only, but I'm very much willing to add Python bindings later, assuming this will be accepted. Everything *outside* `src/cqasm` is binary and build-system-compatible with what's on the develop branch, but uses the newly added API underneath, instead of the now-defunct original flex/bison grammar. That way, going forward, only the new API needs to be updated to add to or change the cQASM language.

The new grammar is *almost* entirely backward-compatible. The only breaking changes that I'm aware of are:

 - A newline (or the newly introduced semicolon) is now required after every operation. I considered the fact that the current parser accepts some (!) operations without preceding newline to be unintentional. I'm pretty sure the grammar would become quite ambiguous if I'd change this.

 - Only the U operation accepts a unitary matrix now. I considered the fact that you could for instance write `X q[1], [1, 0, 0, 0, 0, 0, 1, 0]` to get an "X" gate with an identity matrix to be a bug. Likewise, you could specify a U operation *without* a matrix (not sure if there was a semantic check preventing this, though), which is now also illegal. This is really easy to change though, if needed for some reason.

 - You can now only specify one error model or no error model at all. I considered the fact that you could specify it multiple times (with only the last being recorded) to be a bug. Easy to change if needed, because it's just a semantic check.

 - (Qu)bit index lists may no longer contain duplicates. These seem to have been silently replaced by way of sort + duplicate removal, instead. The sort in particular seems like it might cause confusion when multiple indices are used in conjunction with multi-qubit gates; for example `cnot q[1,2], q[4,3]` would behave like `cnot q[1], q[3] | cnot q[2], q[4]` versus what you'd naturally expect. I had to change a few test cases for this, because they did exactly this a few times. Note that the sorting behavior remains in the old API, as I didn't change its semantic pass.

 - Something like `cnot q[1,2], q[3,4,5]` is now also illegal, because it makes no logical sense. I gave `measure_parity` special treatment here because I think the multi-index syntax is used differently there. Easy to change if needed because the same flag that disables it for `measure_parity` can be added to all the other instructions as well.

 - Bit index lists can no longer access bits beyond the number of qubits. To my knowledge of all current implementations each qubit has exactly one measurement bit attached to it, and there's no point in having more bits when the only classical operation is `not`. This was a natural result of the way I implemented (qu)bit references now, so this'd be rather annoying to change.

 - The depolarizing_channel error model grammar accepted *any* number of floating point arguments. It's now (arbitrarily) limited to 50, which seems like it'd be high enough to me... This is due to overload resolution not supporting variadics, because that would be a completely ridiculous thing to implement for just the freaking error model.

 - The version number was a float, and would thus accept silly things like "version 1e-10". It is now a period-separated list of one or more nonnegative integers, as version numbers normally are. The version number is nevertheless converted to a `double` for the current API for backward compatibility, using only the first two components, with the second component clamped to 9 or less.

It supports a lot of new stuff that I didn't bother removing from my cQASM 2.0 grammar, though:

 - Comments and whitespace are allowed *everywhere*, instead of legal places being exceptions to the rule, as they should be in a sane language. I added the following constructs in addition:
    - C-style block comments `/* comment... */`;
    - Any place in the code where a newline is required can be replaced with a semicolon, more or less like Python;
    - Wherever a newline is *not* allowed syntactically, you can add one by "escaping" the newline with a backslash at the end, also like Python and most other languages.

 - Pretty much everything is a legal name for mappings now that `qubits`, `map`, and `reset-averaging` are the only keywords (#56). You could even remap `q` and `b` if you want, as internally they're now just mappings created implicitly by the `qubits` statement. Mappings can furthermore be any expression, so you could for instance `map 0.707, sqrt2`. You can now also use the syntax `map sqrt2 = 0.707` for the same thing, which I find much more easy to read personally.

 - There is now a much more intuitive notation for U matrices. Instead of the `[a, b, c, d, e, f, g, h]` notation there is now (which still works), you can also use `[a + im*b, c + im*d; e + im*f, g + im*h]`. Note the semicolon in the middle (which can also be a newline) to separate the rows, and that newlines after the `[` and before the `]` are also explicitly legal for readability. Those imaginary number expressions are furthermore *not* hardcoded in the grammar; the new parser understands a lot of expressions involving real and complex numbers, using `im` as a predefined mapping to the imaginary unit, including functions like `sin()`, `sqrt()` etc., and `**` as exponentiation operator, like Python. The guts of the type & value system are capable of much more, all the way down to matrix arithmetic and dynamic expressions once classical logic is added, but so far I only bothered with `+ - * / **` and a bunch of functions on reals and complex numbers.

 - For what it's worth, string literals (used only for the `load_state` instruction right now) are now proper strings supporting escape sequences.

 - All error messages include filename/line/column ranges.

That's just the stuff that translates back into the current API. There's more when you're using the new API:

 - As was the initial goal, the instruction set can be completely configured at runtime, and still be automatically typechecked by libqasm. Semantic analysis in this respect can also be disabled entirely. Same thing applies for error models.

 - Instructions, bundles, subcircuits, and mappings accept annotations, similar to [DQCsim's arbs](https://qe-lab.github.io/dqcsim/intro/arbs.html). They serve the same purpose as pragmas: allowing users to add extra data to stuff not defined in the language itself. That way, libqasm doesn't immediately need to be forked or changed when someone needs some unforeseen feature. Annotations can accept parameters of any type supported by libqasm, as well as raw JSON arguments by using `{|` and `|}` instead of the regular braces for the outermost object (newlines and strings are handled correctly even if they embed an `|}`). One might for instance want to annotate qubits with error model parameters using a mapping or annotate instructions with timing information or waveform indices.

 - Initial mappings and custom functions for use in constant propagation can be added at runtime.

There's some improvements not related to the cQASM language as well:

 - The new API is fully reentrant; no more global state (#94).

 - Multiple errors can be detected and reported at once through recovery mechanisms in the parser and semantic analyzer.

 - Using the new API, the user has access to both a raw abstract syntax tree representation of the file and a semantic tree, the latter being similar to the one in the current API. Both can be debug-dumped in a readable way their entirety, among lots of other things. Refer to [src/cqasm/tests/tutorial.cpp](https://github.com/jvanstraten/libqasm/blob/3f959b0182aa919ad946c941cc35e23bc3e98f54/src/cqasm/tests/tutorial.cpp) for more info.

 - The CMakeLists.txt used to build the new API can be included by other projects using `add_subdirectory` without everything breaking or tests being s l o w l y built and run. The modified version of the current API [does just that](https://github.com/jvanstraten/libqasm/blob/3f959b0182aa919ad946c941cc35e23bc3e98f54/src/library/CMakeLists.txt#L33) instead of including flex/bison and linking against that.

 - I tested (and fixed) the build of everything up to and including the Python module for the following environments:
    - my own OpenSUSE box with a bunch of stuff installed;
    - a bare CentOS 6 + devtoolset 7 box, for which the build system *automatically* downloads and compiled flex and bison (it will try to do this whenever it can't find them or the found versions are too old, but it probably only works on Linux)
    - Windows with MSVC community 2019 according to the OpenQL notes for Windows users, in an otherwise cleanly installed Win10 VM;
    - the above but with MSVC community 2015 (unfortunately my tests don't compile there due to a dependency, but the old API and its tests do work);
    - MacOS (not sure about the version, asked a colleage).
   I would have tried MinGW, but I couldn't for the life of me figure out how to install that beast.

 - The library name for the new API is libcqasm.so instead of lexgram.so. The current name has always bothered me immensely, because shared objects are normally installed directly into `/usr/lib` on Linux (everything else is not portable), and it's impossible to determine what it is based on just that name.

 - Misc. bugfixes that aren't worth mentioning because this is already way too long.

If accepted and time is allocated, I'd like to do the following things in addition:

 - Add auto-generated documentation for the C++ API of some kind.

 - Implement Python bindings (+ documentation) for the new API.

 - Fix setup.py. In its current state, there's no way you could build wheels with it, send them to PyPI, and expect them to work everywhere, due to for instance rpath issues. But I don't see a reason why setuptools wouldn't be able to manage the build for this project, in which case you *should* get all of that for free. I'd probably rewrite it while working on the above bindings.

 - Implement pretty-printing support, to go from a semantic tree/AST back to a valid cQASM file. The new tree system is already written to be easy to modify or build from scratch, so it makes sense for libqasm to be able to go both ways.


jvanstraten:
It was pointed out to me today that any sort of review process is kind of difficult without me first getting documentation sorted out, so I'm working on that first. Note that I'm writing the documenting as if this is already merged. That's not to imply that I'm not open to suggestions; in fact outside of this I haven't really talked to anyone outside of Hans and Razvan yet.

jvanstraten:
Building QX with the updated library works out of the box on my Linux machine, after 964fb99 (I'm not sure what's up with that test case, but the various platforms/compilers seem to disagree what's allowed in the CHECK_THROWS macro). It seems to run just fine for as far as I tested it manually, but it has no automated test suite so I can't do a proper conformance test.

jvanstraten:
Current OpenQL develop builds with e6b4ccc after [these changes](https://github.com/QE-Lab/OpenQL/commit/e494604c06e55199c3eeda599b7450cde732fc34) to its CMakeLists.txt files. I figured it didn't make much sense to update OpenQL's CMakeLists with something that relies on libqasm's directory structure etc. yet again, so I first added a CMakeLists.txt at the root of the repository that normally just links to the new API CMakeLists, but can also include the compatibility headers and the original C++ tests in a decent way.

jvanstraten:
C++ documentation is now available [here](https://libqasm.readthedocs.io/en/latest/doxy/index.html). Once merged I'll remove that readthedocs project so the URL can hopefully be reclaimed for this repository, I kind of expected my username to be in the URL somewhere but nope.

jvanstraten:
As far as Hans, Razvan, and I are concerned, the branch is ready to be merged now. We're still waiting on Richard Versluis and maybe Wouter Vlothuizen to provide comments.

src/cqasm/tests/parsing.cpp
```diff
@@ -0,0 +1,210 @@
...
+        analyzer.register_instruction("measure_y", "Q", false);
+        analyzer.register_instruction("measure_z", "Q", false);
+        analyzer.register_instruction("rx", "Qr");
+        analyzer.register_instruction("ry", "Qr");
+        analyzer.register_instruction("rz", "Qr");
+        analyzer.register_instruction("cnot", "QQ");
+        analyzer.register_instruction("cz", "QQ");
+        analyzer.register_instruction("swap", "QQ");
+        analyzer.register_instruction("cr", "QQr");
+        analyzer.register_instruction("crk", "QQi");
```
razvnane, line 116: Is the c-GATE construct still supported for all possible gates ?

jvanstraten: Short answer: yes to both. Unless I messed up with some or other corner case, everything that was legal and now isn't is explicitly mentioned in my first comment.

Long answer:
 - c- was in fact not supported grammatically for all gates, ex. c-display or c-reset-averaging or c-skip were illegal. Therefore, register_instruction() has a flag you can set to make it illegal, used by the default gateset logic to make the new behavior match the original grammar exactly. In case you're wondering where the c- support comes from in the code (since c- variants of names are never registered anywhere); `c-` is a keyword of a sort that grammatically changes the way an instruction is parsed. I had to do that to not conflict with subtraction expressions.
 - SGMQ is of course still supported, but has changed slightly by way of additional checks, because the original implementation was doing weird stuff that didn't match any of the cQASM documentation. See bullets 4 and 5 of the incompatibilities list in the first comment. IIRC there were a few cases where multiple qubits were not originally supported, in which case I'd have copied the original behavior, but I'm failing to come up with examples so I may be misremembering.

PR #107, opened 2020-07-29
setup.py cleanup and continuous integration

jvanstraten:
This PR copies over the work done to try to do things right for OpenQL's build system into libqasm. Specifically, setup.py is set up in the proper way (extending distutils/setuptools steps for the build process rather than just slamming it in front of `setup()`), a conda recipe is added, and continuous integration using GitHub Actions takes care of testing stuff across the board and publishing wheels to PyPI and Anaconda Cloud (see [RELEASE.md](https://github.com/jvanstraten/libqasm/blob/d376d1b0da47fc66cc2f9cb4a02ac34b124b94a8/RELEASE.md)). The latter requires some secrets that need to be added by a repository admin:

 - `ANACONDA_USER` and `ANACONDA_PASSWORD`: Anaconda cloud username for the account to host `libqasm`.
 - `PYPI_TOKEN`: a [PyPI token](https://pypi.org/help/#apitoken) for the account to host `libQasm`.

The difference in package names is because it never should have contained a capital letter per the Python style guides, and Anaconda enforces that by disallowing capital letters. Once installed, it's `import libQasm` either way.

I tried to enable Actions for this PR already, but that requires push access. See https://github.com/jvanstraten/libqasm/pull/2 for what it would look like. It should turn on on its own once this PR is merged.

Aside from `cleanme.py` being replaced by `setup.py clean`, I don't think this should affect anyone's workflow. This is just a quality-of-life thing for development.


RELEASE.md
```diff
@@ -0,0 +1,34 @@
+Release procedure
+=================
+
+Most of the release process is managed by GitHub Actions. If you follow the
+procedure below, it will automatically test and build all wheels and conda
+packages for all platform, and publish them to PyPI and conda assuming the
+secrets are configured correctly.
+
+ - Make a branch with a name starting with "release" based upon the commit that
```
razvnane, line 9: Please give an example, i.e., release-BRANCHNAME or release/BRANCHNAME, or both?

RELEASE.md
```diff
@@ -0,0 +1,34 @@
...
+   is to be released.
+
+ - Change the version in `src/version.h` (everything else that needs a version
+   should derive from this) and change any other files where applicable
+   (changelog, etc), then commit and make a PR for it.
```
razvnane, line 14: shouldn't we have an exhaustive list here so that all the files that contain the version and need to be changed manually is listed here to prevent inconsistencies?

RELEASE.md
```diff
@@ -0,0 +1,34 @@
...
+   to the GitHub Actions build artifacts only. You can then test these yourself
+   if you have reason to believe that something might be wrong with them that
+   CI might not catch. To find them, go to Actions -> Assets workflow ->
+   click the run for your branch -> Artifacts.
+
+ - If the test or assets workflows fail, fix it before merging the PR (of
+   course).
+
+ - Once CI is green, merge the PR and delete the branch. Instead, push a
+   tag with the appropriate version (customarily `x.y.z` without a `v` in
```
razvnane, line 27: I find 'Instead' confusing because I read it like "don't merge PR and delete the branch".

python/compat/fix-swig-cmdline.py
```diff
@@ -0,0 +1,21 @@
+
```
razvnane, line 1: A comment explaining what problem is fixed with this code would be nice.

RELEASE.md
jvanstraten: This is actually a copypaste mistake from OpenQL, because version.h doesn't exist here. It's only setup.py in this case. But what that means is that there is only one place with the version hardcoded, and everything else reads from that place in some way, so the version numbers are consistent by design.

razvnane (approved):

RELEASE.md
jvanstraten: I don't think slashes are allowed in branch names, but other than that it doesn't matter, CI just matches "release*" in the branch name. But sure, I'll add an example.

RELEASE.md
jvanstraten: Uh, that's fair, I'm not sure what happened there.

python/compat/fix-swig-cmdline.py
jvanstraten: That would be [here](https://github.com/jvanstraten/libqasm/blob/828456d8c918accd9f3a9e7fa74651131dd6d668/python/CMakeLists.txt#L77), where the compat scripts are used. tl;dr it's complicated, features fundamentally needed to have CMake figure out include directory dependencies based on targets and reflect those in the command line for swig are missing prior to 3.12, so I have to do some postprocessing. From 3.12 onwards CMake's own UseSWIG implementation is good enough to handle this.

PR #126, opened 2021-05-31
Fix/sds_133/parsing_c_cr

QFer:
* Fixed parsing binary controlled cr-gate with multiple-qubits and float parameter (angle)
* added test to example3
* version from 0.0.1 to 0.0.2


jvanstraten:
While the changes look like they're correct for this ancient version of cQASM, `src/library/grammar.y` no longer even exists, and the new parser never had the problem that this PR attempts to fix. Also, the version is currently already 0.3. So there's no point in merging this, at least not with develop.

PR #135, opened 2021-09-15
Added tests for barriers instruction

rknegjens:
Added unit tests for the barriers instruction that was added to cQASM v1.0 default instruction set.

Note that we currently test for `barrier q[0], q[1]` failing, as it should given its current implementation using a single qubit gate (type param `Q`). However, the original requirement was for also this syntax to work, although `barrier q[0,1]` is recommened/preferred.

### DIFF
--- a/src/tests/cpp/barriers.cpp
+++ b/src/tests/cpp/barriers.cpp
@@ -1,0 +1,32 @@
+/** This test is for barriers qasm file in the paper **/
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+
+// [[noreturn]] completely breaks MSVC 2015, and is basically unnecessary
+#ifdef _MSC_VER
+#define DOCTEST_NORETURN
+#endif
+
+#include <iostream>
+#include <vector>
+#include <string>
+#include "qasm_semantic.hpp"
+#include "doctest/doctest.h"
+
+TEST_CASE("Test for the barriers.qasm file")
+{
+    #if YYDEBUG == 1
+    extern int yydebug;
+    yydebug = 1;
+    #endif
+
+    // open a file handle to a particular file:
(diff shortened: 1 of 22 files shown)


rknegjens:
Not sure yet why the `Conda (macos-latest)` test is failing.. probably unrelated to this PR

src/tests/cpp/barriers.cpp
```diff
@@ -0,0 +1,32 @@
+/** This test is for barriers qasm file in the paper **/
```
QFer, line 1: In the paper, I think the paper about cqasm? cQasm 1.0 didn't mention barrier so it is not in the paper.

src/tests/test_data/barriers.qasm
```diff
@@ -0,0 +1,10 @@
+version 1.0
```
QFer, line 1: I suggest to add a test to the python test set test_cqasm.py for this barrier.qasm file.

QFer (changes requested):

src/tests/cpp/barriers.cpp
rknegjens: Ah this was result of a hasty search and replace.. indeed there are no barrier examples in the original paper. Will fix this comment

src/tests/test_data/barriers.qasm
rknegjens: Yes good idea, will do

rknegjens:
I disabled the Conda (macos-latest) environment test, due to hard-to-debug issue. The Windows conda tests were already disabled due to issues with file locations.

Note that MacOS and Windows are still being tests for both C++ and Python outside of a Conda environment.

rknegjens:
@QFer could you please re-review this MR? It needs one approval to be merge.d
Thanks!
Rob

QFer (approved):

PR #139, opened 2022-02-15
Fix CXX requirements to require C++11

koffie:
Before this change I got errors like:

```
/code/qutech/libqasm/src/cqasm/tree-gen/include/tree-base.hpp.inc:1198:48: error: expected ';' at end of declaration
    Many<typename std::remove_const<T>::type> c{};
                                               ^
                                               ;
/code/qutech/libqasm/src/cqasm/tree-gen/include/tree-base.hpp.inc:1199:10: warning: 'auto' type specifier is a C++11 extension [-Wc++11-extensions]
    for (auto &sptr : this->vec) {
         ^

```
When building on my machine. It appears that this library was written against the C++11 version of the C++ standard.

This can be fixed by defining the required C++ version in the CMAKE file.

It might have build fine on other machines that already use C++11 by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,9 @@ endif()
 
 project(cqasm CXX)
 
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
+
 # Library type option. Default is a static library.
 option(
     BUILD_SHARED_LIBS
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,7 @@ project(cqasm CXX)
 
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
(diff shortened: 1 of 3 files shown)


wvlothuizen (commented):
Looks good to me, and similar settings are present in OpenQL (which includes libqasm as subdirectory).

OpenQL also adds the following line, and it seems to makes sense to do the same here:
`set(CMAKE_CXX_EXTENSIONS OFF)`

koffie:
> set(CMAKE_CXX_EXTENSIONS OFF)

Indeed adding this makes sense. My main focus with this commit was getting it to also work on my machine. However this extra option ensure more portability (at least if it still compiles). I just verified everything works fine with this addition.

tomrijnbeek (commented):
Looks pretty uncontroversial to me. Could you please rebase this based on the latest develop commit to ensure this change still works at HEAD?

koffie:
I created #144 with the rebase on develop since I don't like force pushing to public branches.

PR #140, opened 2022-02-28
Create commandline tool for quickly invoking the libqasm analyzer from the command line

koffie:


elenbaasc:
When I run [this](https://github.com/QuTech-Delft/libqasm/blob/develop/src/cqasm/tests/v1-parsing/semantic/for-ok-1/input.cq) example, I get the following error: `$ ['Error at input.cq:6:13..14: dynamic values are not supported here']` . Is this to be expected?

Nevertheless, this is a neat tool to check cqasm files.

koffie:
Hmmm no this is not was is expected, since the C++ tests say that it is ok. I will try to see what the difference is between C++ tests and this tool. I did see that there are options to change the configuration of the analyzer. So it might also mean that we need to look more into which settings for libqasm are actually being used in openql and friends.

PR #142, opened 2022-03-09
Fix the Anacoda badge url

abousias:
https://github.com/QuTech-Delft/libqasm/pull/141#discussion_r822447129


koffie (approved):
I'm happy with this change. Although @QFer also thought we could remove it al together until it is fixed.

abousias:
> I'm happy with this change. Although @QFer also thought we could remove it al together until it is fixed.

I didn't get any other signal for this. So for now it is ok to merge.

PR #148, opened 2022-09-16
Fix Bison building old 3.0 version

pablolh:
With this commit, if Bison is not preinstalled on the system, version 3.8 will be built, instead of too old version 3.0, which doesn't build with newer glibc.


jvanstraten (approved):
LGTM

Issue #151, opened 2023-06-12
Memory leak in src/library/libQasm.hpp


rturrado:
Fer Grooteman reported a memory leak on libqasm, and pointed us to the code he thought this memory leak was coming from, a `new`ed memory in `src/library/libQasm.hpp` that was never released.

rturrado:
Update:
- Created a libqasm branch fix_issue_151 in order to work on this issue.
- Fixed the original issue.
- Added ASAN compilation and fixed some more memory leaks.
- TODO: fix test-v1-parsing memory leaks.

rturrado:
Closing as fixed.

PR #153, opened 2023-07-04
DRAFT: Fix memory leak for QI

QFer:
* Solved memory leaks for huge cqasm files as encountered in QI (issue #151)
* changed code where objects were copied (without copy constructors) invalidating the pointers
* Added destructors to delete memory for vectors with pointer types
* Changed vector of <SubCircuit> to <SubCircuit *>
* Free memory when calling parse_file twice
* Changed some test code to correctly use the objects from the library

### DIFF
--- a/python/libQasm.i
+++ b/python/libQasm.i
@@ -46,7 +46,7 @@ namespace std {
 %include "cqasm-py.hpp"
 
 namespace std {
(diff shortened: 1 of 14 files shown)


QFer:
Fixed in https://github.com/QuTech-Delft/libqasm/issues/151

PR #154, opened 2023-07-05
Fix issue 151

rturrado:
This commit:
- adds the Address Sanitizer flag (ASAN_ENABLED) to the Debug builds, so that memory issues are raised when running tests.
- fixes the original leak detected by @QFer (in `libQasm.hpp`).
- fixes some more leaks detected by the Address Sanitizer (in `qasm_ast.hpp`, `qasm_new_to_old.hpp`, and `cqasm_v1_parser.y`).


.github/workflows/test.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       if: matrix.os == 'windows-latest'
       run: choco install winflexbison3 --version 2.5.18.20190508
     - name: Configure
-      run: cmake . -DCMAKE_BUILD_TYPE=Debug -DLIBQASM_BUILD_TESTS=ON -DBUILD_SHARED_LIBS=OFF
+      run: cmake . -DCMAKE_BUILD_TYPE=Debug -DLIBQASM_BUILD_TESTS=ON -DBUILD_SHARED_LIBS=OFF -DASAN_ENABLED=ON
```
pablolh, line 42: How much slower is it?
I read online:
"Address Sanitizer has significant performance and memory costs. Expect a system that is 50-100% slower and needs a lot of RAM."

You're enabling for all three platforms, maybe one is enough.

CMakeLists.txt
```diff
@@ -77,6 +77,20 @@ endif()
 set(LIBQASM_GENERATED_PYTHON_FILES ${CMAKE_CURRENT_BINARY_DIR}/py_gen)
 file(MAKE_DIRECTORY ${LIBQASM_GENERATED_PYTHON_FILES})
 
+# Address sanitizer
+if(ASAN_ENABLED)
+    string(REGEX REPLACE "/RTC(su|[1su])" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
```
pablolh, line 82: If this is only for MSVC, might be worth it placing it into the MSVC if section below

src/library/libQasm.hpp
```diff
@@ -2,54 +2,47 @@
...
 
-        const compiler::QasmRepresentation& getQasmRepresentation() const
-        {
-            return qasm_rep_;
+class libQasm {
+public:
+    libQasm() : parse_result_{ 0 }, qasm_rep_{} {}
+
+    void parse_string(const std::string &qasm_str) {
+        compiler::QasmSemanticChecker sm(qasm_str);
```
pablolh, line 17: Much better now!

.github/workflows/test.yml
rturrado: It is much slower, but I would also be more comfortable leaving the Address Sanitizer on for all the platforms. Notice that the Address Sanitizer is only enabled for a GitHub Actions job, not for regular builds.

As of numbers, we can consider these two runs below. I'm not 100% sure we are running exactly the same tests because they are different libqasm branches, and we don't have logs for libqasm runs without Address Sanitizer enabled. But they can give us an idea:
- [With](https://github.com/QuTech-Delft/libqasm/actions/runs/5476046943) Address Sanitizer enabled: tests take 4.30s for [Ubuntu](https://github.com/QuTech-Delft/libqasm/actions/runs/5476046943/jobs/9973018935) (3.3% of the total run), 7.92s for [MacOS](https://github.com/QuTech-Delft/libqasm/actions/runs/5476046943/jobs/9973019182) (3.8%) and 29.16% for [Windows](https://github.com/QuTech-Delft/libqasm/actions/runs/5476046943/jobs/9973019412) (11.5%).
- [Without](https://github.com/QuTech-Delft/libqasm/actions/runs/5500737691) Address Sanitizer enabled: [Ubuntu](https://github.com/QuTech-Delft/libqasm/actions/runs/5500737691/jobs/10023850757) 1.42s, [MacOS](https://github.com/QuTech-Delft/libqasm/actions/runs/5500737691/jobs/10023850828) 4.22s, [Windows](https://github.com/QuTech-Delft/libqasm/actions/runs/5500737691/jobs/10023850875) 5.87s.

CMakeLists.txt
rturrado: That's right, yes. Much better.

src/cqasm/src/cqasm-v1-parser.y
```diff
@@ -1,3 +1,15 @@
+%{
+#include <stdlib.h>
+#include <string.h>
+
+/*  handy clean-up function */
+void free_string(char *s) {
+    if (s != NULL && strcmp(s, "") != 0) {
```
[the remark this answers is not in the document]
rturrado: I think so, yes. Thanks so much for pointing this out. Even not being 100% sure, I think that second part of the check makes the code less easier to read, more complicated, and more unreliable (for the sake of some corner case performance?).

.github/workflows/test.yml
pablolh: Fair enough since in all cases the runtime of tests is still short.

pablolh (approved):

PR #155, opened 2023-07-12
v3

rturrado:
Main changes:
- Fixed GitHub Actions' test.yml workflow.
- Added `v1.0` and `v1` directories to `include`, `src`, and `test` directories. And to `build/include` and `build/src` directories.
  - In order to hold current v1.0 and v1 files (cqasm 1.1 and 1.2).
  - And in order to allow for a future inclussion of v3 files.
  - Most of the changes in this PR come from adapting to this new folder structure. For example:
    | **develop** | **v3 (PR)** |
    |-|-|
    | cqasm-v1-parser-helper.hpp | v1/cqasm-parser-helper.hpp |
    | cqasm-v1-types-gen.hpp | v1/cqasm-types-gen.hpp |
- Reorganized folders:
    | **develop** | **v3 (PR)** |
    |-|-|
    | src/cqasm/cmake | cmake | 
    | doc/source | doc/src | 
    | doc/old | doc/v1.0 | 
    | src/cqasm/include | include | 
    | src/library | include/v1.0 | 
    | src/cqasm/include/\*-v1-\*.hpp | include/v1/*.hpp | 
    | src/cqasm/tests/v1-parsing | res/v1/parsing | 
    | src/cqasm/tests/grover.cq | res/v1/grover.cq | 
    | src/tests/test_data | res/v1.0 | 
    | src/cqasm/src | src | 
    | src/cqasm/func-gen | src/func-gen | 
    | src/cqasm/src/\*-v1-\*.cpp | src/v1/*.cpp | 
    | src/cqasm/tests | test/v1 | 
    | src/tests/cpp | test/v1.0/cpp | 
    | src/tests/python | test/v1.0/python | 
- Test:
  - We only have one test folder.
  - We only use `googletest`, not `doctest`.
  - We only have one test executable, not one executable per test.
  - Added a `res` folder to contain all the resource files.
- Updated to use CMake 3.12.
  - TODO: I think this was a Conan requirement, but I'm not completely sure why.
- Updated to use C++20.
  - We could move to C++23, but we also intend to do a libqasm Conan package, and Conan still doesn't support C++23.
- Removed git submodules.
  - There were only 2 git submodules: `doctest`, which is no longer used, and `tree-gen`, that is now retrieved via `FetchContent`.
    - `FetchContent` is more user transparent (you don't have to remember to add `--recurse-submodules` to `git clone`).
    - It also separates the library dependencies from the source code, since dependencies are downloaded to the binary folder.
    - And, in case you use Conan as package manager, it doesn't force you to clone the dependencies source. That is not the case here, because there is no `tree-gen` package, but it is for OpenQL.

Conan:
- `conanfile.py`.
- `include/test_libqasm_conan_package.hpp`.
- `test_package`.

Refactoring:
- `include/cqasm-version.hpp`, `src/cqasm-version.cpp`, `test/cqasm-version.cpp`:
  - Used the class `ParseHelper` as an example of a refactoring that could be used elsewhere in the code.
  - Simplified 3 reading functions that were doing basically the same, repeating code, into a single constructor.
  - Moved specific code out to a class, e.g. `ScannerFlexBison`.
  - Added dependency injection to make the class testable, so `ParseHelper` now receives a pointer to an scanner adaptor, which can be flex/bison or whatever.
  - Added `test/cqasm-version.cpp` tests.

Late removals:
- Removed the files related to the libqasm branch without flex/bison dependencies, because they don't really belong here. They're now in a separate branch, `v3_develop`.
- Removed all the references to v3. I had v3 folders at `include`, `src`, and `test`, but they were not doing anything really, except for serving as an example of how to separate v1 code from v3 code.

Specific changes:
- `.github/workflows/test.yml`:
  - Updated versions of actions.
  - Renamed `cpp` job to `cpp_cmake` and added `cpp_conan`.
- `include/cqasm.hpp`:
  - There is an important change here.
    - Before, in `src/cqasm/include/cqasm.hpp`, there was only a `cqasm` namespace, which was `using namespace v1`. So all `v1` contents were available just via the `cqasm` namespace.
    - Now, we just include `v1/cqasm.hpp`. And we leave open the possibility to include a `v3/cqasm.hpp` in the future. The source code will then have to use namespace `cqasm::v1` or `cqasm::v3`.
- `include/cqasm-string-builder.hpp`:
  - Moved `StringBuilder` class out of `src/cqasm/include/cqasm-v1-ast.hpp`, as it is not v1 specific, and placed it in `include`, close to other common code such as `cqasm-annotations.hpp` or `cqasm-error.hpp`.
- `python`:
  - `CMakeLists.txt`: updated to use C++23. This is not going to be part of the Conan package, so it can be updated to C++23.
  - Changes to consider new folder structure.
- `include/{v1,v1.0}/README.md`:
  - Updated.
- `Dockerfile`:
  - Changes to consider new folder structure.
- `pyproject.toml`:
  - Changes to consider new folder structure.
- `README.md`:
  - Updated.


doc/v1.0/qtools/Makefile
```diff
@@ -1,113 +1,113 @@
-# Makefile
```
pablolh, line 1: I'm wondering if the "old/" in the file path indicates that it's some old code/docs for cQasm 1.0.
Maybe keep it, personally I'm not opposed to removing it completely.

doc/v1.0/qtools/gstartfollow.c
```diff
@@ -1,20 +1,20 @@
-/* gstartstop.c */
```
pablolh, line 1: don't understand why this shows up in the diff, files are identical?
Could it be line endings that changed?

pablolh: What is this qtools? Can we remove it?

res/v1.0/bare_minimum.qasm
pablolh, line 1: Why is there a v1.0 folder AND a v1 folder?

test/main.cpp
```diff
@@ -0,0 +1,16 @@
+#include "v1/parsing.hpp"
+
+#include <gmock/gmock.h>
+#include <iostream>
+
+int main_impl(int argc, char** argv, std::ostream&) {
+    ::testing::InitGoogleMock(&argc, argv);
+
+    register_v1_parsing_tests();
```
pablolh, line 9: I think you don't need this file if you link against the gtest CMake target that provides a main, see 
http://google.github.io/googletest/quickstart-cmake.html

So, maybe it works if you use target_link_libraries(XXX gtest_main)

test/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+# Test sources
+set(test_sources
+    ${v1_0_test_sources}
+    ${v1_test_sources}
+)
+set(app_sources
+    "${CMAKE_CURRENT_SOURCE_DIR}/cqasm-version.cpp"
+    "${CMAKE_CURRENT_SOURCE_DIR}/main.cpp"
+)
+list(APPEND app_sources ${test_sources})
```
pablolh, line 25: Instead of having those slightly awkward PARENT_SCOPE and list(APPEND ...)
Can you use repeated calls to target_sources, see https://cmake.org/cmake/help/latest/command/target_sources.html ?
That would mean declaring the test executable before add_subdirectory(v1)
and then in the v1 subfolder, you call target_sources with the sources that are there to append them to your test executable, without PARENT_SCOPE.

test/cqasm-version.cpp
```diff
@@ -0,0 +1,224 @@
...
+#include <gmock/gmock.h>
+#include <memory>
+
+using namespace ::testing;
+using namespace cqasm::error;
+using namespace cqasm::version;
+
+
+TEST(parse_file, filename_does_not_exist) {
+    EXPECT_THROW(parse_file("res/cqasm_version/does_not_exist.cq"), AnalysisError);
```
pablolh, line 14: Testing seems a bit incomplete to me here because there is no further cQasm in your files.

I suggest adding some tests like those:
- one has no version but some cQasm instruction, so it would start with e.g. "h q[0]", this should fail
- one calls a gate called "version", for instance "version q[1:6], 3.14159; measure_all", should fail
- one has a normal version 1.0 plus some instructions (should succeed).

test/cqasm-version.cpp
```diff
@@ -0,0 +1,224 @@
...
+    const char *filename{ "res/cqasm_version/version_1_1_abc.cq" };
+    EXPECT_THROW(parse_string("version 1.1.abc", filename), AnalysisError);
+}
+TEST(parse_string, string_version_1_0) {
+    const char *filename{ "res/cqasm_version/version_1_0.cq" };
+    EXPECT_EQ(parse_string("version 1.0", filename), Version{ "1.0" });
+}
+TEST(parse_string, string_version_1_1) {
+    const char *filename{ "res/cqasm_version/version_1_1.cq" };
+    EXPECT_EQ(parse_string("version 1.1", filename), Version{ "1.1" });
```
pablolh, line 155: Why does parse_string takes a filename as second argument?

test/v1/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+set(v1_test_sources
+    "${CMAKE_CURRENT_SOURCE_DIR}/parsing.cpp"
+    "${CMAKE_CURRENT_SOURCE_DIR}/tutorial.cpp"
+    PARENT_SCOPE
```
pablolh, line 4: Here again I think it's possible to use target_sources
https://stackoverflow.com/questions/8934295/add-source-in-a-subdirectory-to-a-cmake-project

conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "0.5.6":
+    url: "https://github.com/QuTech-Delft/libqasm/archive/refs/tags/v3-0.5.6.zip"
```
pablolh, line 3: what is this 0.5.6 version?
I see "v3" in the url, does that refer to the name of the branch you want to merge?

pablolh: Do we need to modify this once we merge?

conanfile.py
```diff
@@ -0,0 +1,115 @@
...
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "asan_enabled": [True, False],
+        "build_python": [True, False],
+        "build_tests": [True, False],
+        "compat": [True, False],
+        "tree_gen_build_tests": [True, False]
+    }
+    default_options = {
```
pablolh, line 32: I like this much better than raw CMAKE. Is this duplicated somewhere or is this file the only source of truth for compilation now?

pablolh (changes requested):
I tried doing a relatively exhaustive review but my browser is super slow because of the large number of files haha

doc/v1.0/qtools/Makefile
rturrado: What I'm not sure either is why these files are appearing in the code review. You may be right in that they contain line ending changes. I will check that.

doc/v1.0/qtools/gstartfollow.c
rturrado: Regarding the first comment, what I said above, it may be line endings, yes. I will check.

Regarding the second comment, I have no idea. I just did file reorganization. I would include that as part of a separate PR, "cleaning libqasm" or something like that.

test/CMakeLists.txt
rturrado: Thanks Pablo, that's also worth having a look.

According to the [add_executable](https://cmake.org/cmake/help/latest/command/add_executable.html#command:add_executable) documentation, you can just define it and add the sources later via `target_sources` as you sayL `New in version 3.11: The source files can be omitted if they are added later using [target_sources()](https://cmake.org/cmake/help/latest/command/target_sources.html#command:target_sources).`

I don't know though if the target name is visible from the subfolders. I will check.

test/cqasm-version.cpp
rturrado: Good! They all look like very sensible tests to do. I will try and add them.

test/v1/CMakeLists.txt
rturrado: Yep, I'll give it a go. Thanks!

conandata.yml
rturrado: Very good spot, because it's a indeed a `v3` reference I missed, and very good questions.

This has to do with having a libqasm Conan package. Since that's a different piece of work, I'll remove all the files related to it. This is a file that has to go in `conan-center-index/libqasm`. I don't think it should belong here anyway.

I take the chance though, to explain how that libqasm Conan package should work. Let's say a client, e.g. OpenQL, request to use libqasm via its `conanfile.py`. The `conandata.yml` file specifies where to download the libqasm sources from. This file should be changed every time you create a new libqasm Conan package, yes. But, as I stated above, not here, but in the `conan-center-index`, another GitHub project. This procedure should be documented. Summarising it here:
- You have a new libqasm commit that you want to make into a release.
- You create a release. This will generate some artifacts, e.g. `https://github.com/QuTech-Delft/libqasm/archive/refs/tags/0.4.0.zip`.
- You update `conan-center-index/libqasm` files, including this `conandata.yml` to point to the new release.
- Clients can update their code to work with the new release, e.g. by modifying the version of the libqasm library in their `conanfile.py` files.

conanfile.py
rturrado: We are still able to use CMake as we've been doing it until now. If you have a look at the GitHub Actions `test` workflow, you'll notice that we build for CMake/Conan and ubuntu/macos/windows.

What I've removed is the use of `git submodules`, because you don't need them with Conan. In case you go via the CMake path, CMake will use FetchContent to retrieve the needed projects.

conandata.yml
rturrado: I've fixed this. I've removed `conandata.yml`, and I've changed `conanfile.py` to retrieve sources from the current location. I've left the `test_package` folder though. You may want to install the library locally, to the Conan cache, for example to have it available for other projects like OpenQL. In that case, you need `test_package` to test the installation. You can build libqasm and install it in the Conan cache by using `conan create`. For example:
```
conan create . --version=1.0 -s:h compiler.cppstd=20 -o libqasm/*:compat=True -b missing
```

doc/v1.0/qtools/gstartfollow.c
rturrado: Same comment as above, I have checked out develop in my computer and compared the two `gstartfollow.c` files, and they are exactly the same. Also the left hand side pane of this code review doesn't mark Makefile as a modified file.

test/main.cpp
rturrado: OK. I've tried this, and, effectively, deleting the `test/main.cpp` file and linking to `gtest_main` worked fine.

However, I have just noticed that we still need the `main.cpp` file to call `register_v1_parsing_tests()`. This is a function in `test/v1/parsing.cpp`, which loops over a directory and calls `::testing::RegisterTest` several times.

test/CMakeLists.txt
rturrado: Fixed. I loved changing this code. It's much nicer now. I removed some lines of code. And I notice the good practices point to using all these `target_*` commands. Thanks for the comment!

test/v1/CMakeLists.txt
rturrado: Already fixed.

rturrado: By the way, I've had a look into using `target_source` for `cqasm-lib-obj`, the library defined in `src/CMakeLists.txt`, but I don't think it's possible there, because it is an [Object library](https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries).

test/cqasm-version.cpp
rturrado: Added!

test/cqasm-version.cpp
pablolh: I see, for the future we should set this filename as some default e.g. "<stdin>" or "<unknown>" or sth more sensible like this, to avoid the confusion.

test/main.cpp
pablolh: Good point
If you still want to go without main, you can use static initialization, see
https://stackoverflow.com/questions/34775254/how-to-auto-register-test-cases-in-a-test-system

Calling TEST_CASE is actually already creating an object with a static member which is initialized by calling the test registration function.
So what you'd need roughly is to create a "dummy" class at the end of e.g. this parsing.cpp, with a static member, and initialize this static member with some return value from register_v1_parsing_tests().
Whether it's better or not I let you judge :)

res/v1.0/bare_minimum.qasm
pablolh: omg, this is confusing. Do you know if there are still usages for the old API?

test/cqasm-version.cpp
pablolh: "unknown" or "stdin"
Github removed my signs

conanfile.py
pablolh: So the compilation options are defined in 2 different places I guess?
What's your vision for the future? we remove CMake?

res/v1.0/bare_minimum.qasm
rturrado: All the v1.0 stuff is included if you compile with `-DLBQASM_COMPAT`. The code in `include/v1.0/` is what we've been changing lately to fix memory leaks (e.g. `libQasm.hpp`, `qasm_new_to_old.hpp`, `qasm_semantic.hpp`). That seems to be part of the SWIG interface. However, it's worth noticing that the CMakeLists.txt in that directory is not used in `develop`. That file at least, I understand, could be erased.

test/main.cpp
rturrado: I think I prefer the `main` implementation, yes. It looks simpler to me beforehand.

test/cqasm-version.cpp
rturrado: Haha. Yes, good point again. I'll have a look into this one!

conanfile.py
rturrado: No, no. The compilation options remain in the CMakeLists.txt file. We don't get rid completely of CMake. Conan wraps around CMake. But, what we do get rid of is package management in CMake.

At the moment, at the beginning of the `src/CMakeLists.txt` file, we are trying to find a series of packages (e.g. `flex`, `bison`, and `m4`):
- In `src/CMakeLists.txt`, there is a line `include(../cmake/flex-bison.cmake)`.
- In `cmake/flex-bison.cmake`, there are a few `find_package` lines (e.g. `find_package(BISON 3.0)`), and then some checks (e.g. `if(NOT BISON_FOUND)`). If a given package isn't found, then it is downloaded, build it, whatever.

I've left this stuff in case we still want to build using a CMake command line (e.g. `cmake .. -B . -DCMAKE_BUILD_TYPE=Debug...`), but, if we only used Conan in the command line, this `cmake` folder, and that `src/CMakeLists.txt` code could be removed completely. You may just want some safety checks in the CMakeLists.txt file (e.g. `find_package(bison REQUIRED)`).

As of plans for the future, if we are comfortable with Conan, and see no major drawbacks using it, I would drop the package management via CMake (thus the possibility of using CMake command lines).

test/cqasm-version.cpp
rturrado: I was a bit sure this was being done at the moment. If you have a look at `include/cqasm_version.hpp`, you'll see that the functions `parse_file(FILE *file, const string &filename="<unknown>")` and `parse_string(const string &data, const string &filename="<unknown>")`, already use an `"<unknown>"` default argument for `filename`.

What I could do is to propagate this to `ParseHelper`, so that the constructor would look like: `ParseHelper(std::unique_ptr<ScannerAdaptor> scanner_up, const std::string &filename="<unknown>")`.

rturrado: Changed `ParseHelper` so that it uses a default `"<unknown>"` filename.

And added a few tests to `parse_file`, `parse_string`, and `ParseHelper` to check everything is OK when no filename is used.

pablolh (dismissed):
Did you check that OpenQL and QX-simulator were still building and testing OK with this change?

pablolh:
Why are there 3 C++ tests builds that don't run?

rturrado:
> > Why are there 3 C++ tests builds that don't run?
> 
> Yes, I've seen that, but I have no idea. I will see if I can remove them.

OK, fixed. Thanks to this [explanation](https://github.com/orgs/community/discussions/26698#discussioncomment-3252954). Since I added a new `cpp_conan` job, I changed the name of the previous `cpp` job to `cpp_cmake`. It turns out that you can state, in the branch protection rules, what tests are required for merging a pull request. In our case, those tests were `cpp` for ubuntu/macos/windows, so GitHub was eternally waiting for them to complete.

I've changed the branch protection rules now to require `cpp_cmake` and `cpp_conan` for ubuntu/macos/windows.

pablolh (dismissed):

pablolh (approved):

PR #158, opened 2023-08-25
Create .readthedocs.yaml

pablolh:
Required to avoid build failures
This file comes from OpenQL.


rturrado (approved):

rturrado:
I'll update the branch protection rules as soon as I get permissions from Olaf, then we can merge this change.

PR #160, opened 2023-09-14
Add Python3.10 and 3.11, remove Python < 3.8, Fix Python API

pablolh:
I need to use the Python API of libqasm, and this is not available for my Python version.

I don't know in fact if it's necessary to do all this, Python3 versions are not so different?
I don't know if all builds are required - it's quite sad to have to do 4 builds for all Python versions for each OS...

### DIFF
--- a/.github/workflows/assets.yml
+++ b/.github/workflows/assets.yml
@@ -15,10 +15,10 @@ jobs:
     strategy:
       matrix:
         python:
-        - '3.6'
-        - '3.7'
         - '3.8'
         - '3.9'
+        - '3.10'
+        - '3.11'
     steps:
     - uses: actions/checkout@v2
       with:
@@ -68,10 +68,10 @@ jobs:
         manylinux:
         - 2014
         cpython_version:
-        - 'cp36-cp36m'
-        - 'cp37-cp37m'
         - 'cp38-cp38'
         - 'cp39-cp39'
+        - 'cp310-cp310'
+        - 'cp311-cp311'
     steps:
     - uses: actions/checkout@v2
       with:
@@ -120,10 +120,10 @@ jobs:
     strategy:
       matrix:
         python:
-        - '3.6'
-        - '3.7'
         - '3.8'
         - '3.9'
+        - '3.10'
+        - '3.11'
     steps:
     - uses: actions/checkout@v2
       with:
@@ -176,10 +176,10 @@ jobs:
         - macos-latest
         #- windows-latest
         python-version:
-        - '3.6'
-        - '3.7'
         - '3.8'
         - '3.9'
+        - '3.10'
+        - '3.11'
     steps:
     - uses: actions/checkout@v2
       with:
--- a/python/conda/meta.yaml
+++ b/python/conda/meta.yaml
@@ -23,6 +23,7 @@ requirements:
     - python
     - plumbum
     - delocate # [osx]
+    - numpy
   run:
     - python
 
@@ -34,6 +35,7 @@ test:
     - test
   requires:
     - pytest
+    - numpy
   commands:
     - pytest test
 
--- a/python/module/cqasm/v1x/__init__.py
+++ b/python/module/cqasm/v1x/__init__.py
@@ -1,4 +1,6 @@
-import libQasm;
+import libQasm
+import cqasm.v1x.ast as ast
+import cqasm.v1x.semantic as semantic
 
 class Analyzer(libQasm.V1xAnalyzer):
 
@@ -6,7 +8,6 @@ class Analyzer(libQasm.V1xAnalyzer):
     def parse_file(*args):
         retval = libQasm.V1xAnalyzer.parse_file(*args)
         if len(retval) == 1:
-            import cqasm.v1x.ast as ast
             return ast.Root.deserialize(retval[0].encode("utf-8", errors="surrogateescape"))
         return list(retval[1:])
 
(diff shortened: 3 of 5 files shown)


.github/workflows/assets.yml
rturrado, line 1: Maybe worth updating the `Manylinux` `cpython` versions too?

rturrado:
> I don't know in fact if it's necessary to do all this, Python3 versions are not so different?

I don't really know.

> I don't know if all builds are required - it's quite sad to have to do 4 builds for all Python versions for each OS...

For `test.yml` we are only testing with the latest Python, 3.11. I wouldn't mind any of the following options:
- Test only with 3.11 for all the OSes.
- Test with 3.8 to 3.11 for all the OSes.
- Test with 3.8 to 3.11 for an OS, e.g. `ubuntu`, and then only with 3.11 for the rest of the OSes.

.github/workflows/assets.yml
pablolh: Absolutely, I forgot

pablolh:
I fixed also the Python API which was broken. SWIG typemap for std::vector<std::string> was not working properly, and the Python API was unusable.
As part of that, I've added a test in Python which also made me familiar with this library.

I will be using the Python libqasm in qx-simulator from now on, to decouple the high-level features like slight optimizations, from the heavy number crunching in Superpositeur.
Because of that, I realized this libqasm Python API was broken and untested.

pablolh:
Ah! I realized finally that those 2 API were coexisting
- v1 had already some Python tests (which failed because I removed typemaps) and I had ignored them
- v1x had no Python tests, so that's what I added, and I placed it now in tests/v1x/python as for the existing v1 Python tests

It seems to me that Quantum Inspire SDK uses v1, but not v1x.
Did you try and look at the differences and if so, do you see good and bad things?
I would rather avoid having two APIs in parallel, it's very confusing. Maybe we need to migrate QI SDK.

I think Jeroen developed v1x later, it probably accounts for new things, and it's maybe explained in the docs. However, for now we don't need matrices and vectors in cQasm.

rturrado:
> I fixed also the Python API which was broken. SWIG typemap for `std::vector<std::string>` was not working properly, and the Python API was unusable. As part of that, I've added a test in Python which also made me familiar with this library.

So `%template(vectors) vector<string>` is not enough and we need a `typemap`?

> I will be using the Python libqasm in qx-simulator from now on, to decouple the high-level features like slight optimizations, from the heavy number crunching in Superpositeur. Because of that, I realized this libqasm Python API was broken and untested.

Great!

rturrado:
> Ah! I realized finally that those 2 API were coexisting
> 
> * v1 had already some Python tests (which failed because I removed typemaps) and I had ignored them
> * v1x had no Python tests, so that's what I added, and I placed it now in tests/v1x/python as for the existing v1 Python tests
> 
> It seems to me that Quantum Inspire SDK uses v1, but not v1x. Did you try and look at the differences and if so, do you see good and bad things? I would rather avoid having two APIs in parallel, it's very confusing. Maybe we need to migrate QI SDK.

My understanding, after having a look at the `include/{v1x,v10}` folders, is that:
- All the parsing code is in `v1x`.
- `v10` seems to have:
  1. connectivity code for when you use the library from Python: e.g. `libQasm.hpp`, and
  2. compatibility layer code.

I also have the impression, as you do, that the compatibility code is always used by the libqasm users.

I guess we could do some work to effectively organize and clarify all this. We may:
- put the `include/v10` folder contents within `include/v1x`, or
- split the `include/v10` folder into a python connection folder and a compatibility layer folder.

There is no `src/v10` folder.

However, we do have different resource files and tests for `v10` and `v1x`. I.e. the distinction seems to be important there.

> I think Jeroen developed v1x later, it probably accounts for new things, and it's maybe explained in the docs. However, for now we don't need matrices and vectors in cQasm.

python/libQasm.i
rturrado, line 1: So `vector<string>` is broken for SWIG? I understand you tried using it in your code, and it wasn't working, right?

python/module/cqasm/v1x/primitives.py
rturrado, line 1: Just curious about how you came to find out you needed this change.

rturrado (commented):
Brilliant work Pablo. I just mentioned a couple of questions. I am mainly concerned about `vector<string>` not working in SWIG, since they provide those `std_vector.i` and `std_string.i`. I have also seen that Python/MacOS and Conda tests are failing.

python/module/cqasm/v1x/primitives.py
pablolh: Pretty-printing was showing strings like gate names starting with b', like b'x'
That's what Python does when you call str() on a bytes object without specifying encoding, see
https://docs.python.org/3/library/stdtypes.html#str

This was also wrong in the AST: because those bytes objects (used for deserialization) were cast to str by the library without specifying encoding.

python/libQasm.i
pablolh: To try it out, I've just pushed the same change with the test, but reverting the changes to libQasm.i, see if the test passes

It declares the std::vector<std::string> SWIG template, but the test fails because even though the vector of strings can be accessed with [], the elements are not Python strings
At least that's what I got, let's see now

pablolh:
I would indeed also rather use the existing SWIG code for vector of strings typemaps, if only it worked :D

pablolh:
Mmmh good thing I tried it, I think it's only on my machine :D
Maybe has to do with SWIG version

pablolh:
Sorry for the email spam btw

rturrado:
> Mmmh good thing I tried it, I think it's only on my machine :D Maybe has to do with SWIG version

Ooops!

rturrado:
> Sorry for the email spam btw

No problem at all!

rturrado:
For the Conda error you *might* need adding an `encoding=`, as in the line below, at several points in the `python/module/cqasm/v1x/__init__.py` file:

```
 return ast.Root.deserialize(retval[0].encode(encoding="utf-8", errors="surrogateescape"))
```

As stated in this [SO answer](https://stackoverflow.com/a/66361497/260313).

PR #166, opened 2023-11-16
tree-gen classes can now dump to JSON

rturrado:
Updated `libqasm` to use the latest commit ID for `tree-gen/json_dumper`.

Updated v3x parsing tests to check the JSON dump.
I have only enabled the checks for one test for the time being.
Since these are file-based checks, enabling them for all the tests may be quite time consuming.


src/v3x/cqasm-py.cpp
```diff
@@ -114,3 +114,21 @@ std::vector<std::string> V3xAnalyzer::analyze_string(
         ? std::vector<std::string>{ tree::base::serialize(result.root) }
         : result.errors;
 }
+
+
+V3xParseResult::V3xParseResult() {
+    parse_result = std::make_unique<cqasm::v3x::parser::ParseResult>();
```
pablolh, line 120: can move this to the header?

pablolh: Also, better to use member initializer list I believe

include/v3x/cqasm-analyzer.hpp
```diff
@@ -8,71 +8,22 @@
...
 #include <functional>
+#include <string>
 
 
 /**
  * Namespace for the \ref cqasm::analyzer::Analyzer "Analyzer" class and support classes.
  */
 namespace cqasm::v3x::analyzer {
 
-/**
```
pablolh, line 25: I like the file splitting

test/cqasm-utils.cpp
```diff
@@ -21,3 +25,80 @@ TEST(equal_case_insensitive, both_contain_lowercase_and_uppercase) {
...
 TEST(equal_case_insensitive, different_strings) { EXPECT_FALSE(equal_case_insensitive("123", "ABC")); }
+
+
+TEST(to_json, v1x_parser_errors) {
+    auto input_file_path = fs::path{"res"}/"v1x"/"parsing"/"grammar"/"expression_recovery"/"input.cq";
+    cqasm::v1x::analyzer::Analyzer analyzer{};
+    auto semantic_ast_result = analyzer.analyze(input_file_path.generic_string());
+    auto json_result = to_json(semantic_ast_result);
+    auto expected_json_result = std::string{
+        R"delim({"errors":["res/v1x/parsing/grammar/expression_recovery/input.cq:4:21: syntax error, unexpected ','""res/v1x/parsing/grammar/expression_recovery/input.cq:4:31: syntax error, unexpected ','""res/v1x/parsing/grammar/expression_recovery/input.cq:4:37: syntax error, unexpected ')', expecting ']'""res/v1x/parsing/grammar/expression_recovery/input.cq:4:61: syntax error, unexpected ','""res/v1x/parsing/grammar/expression_recovery/input.cq:4:67: syntax error, unexpected ')', expecting ',' or ']'""Failed to parse res/v1x/parsing/grammar/expression_recovery/input.cq"]})delim"
```
pablolh, line 36: Is this valid JSON?
It looks like it's missing commas between quotes, why `""`?

python/module/cqasm/v3x/__init__.py
```diff
@@ -1,34 +1,56 @@
...
 
     def analyze_string(self, *args):
         retval = super().analyze_string(*args)
         if len(retval) == 1:
-            import cqasm.v3x.semantic as semantic
             return semantic.Program.deserialize(retval[0].encode("utf-8", errors="surrogateescape"))
         return list(retval[1:])
+
+
+class ParseResult(libQasm.V3xParseResult):
```
pablolh, line 47: It would be beneficial to have a Python test to verify this

pablolh: I think I should have finished and merged my own PR, but feel free to take inspiration from:
https://github.com/QuTech-Delft/libqasm/pull/160/files#diff-e567d5b848a1f3493e90e2b5c7f6739275321a79652474781ec5f43b91514a76
and the whoel PR

src/v3x/cqasm-py.cpp
rturrado: Thanks for the comments, Pablo! I have changed them to use a member initializer list, very good point. However, I prefer to keep them in the source file, so that we don't have to include headers for `ParseResult` and `AnalyzerResult` in the header file.

test/cqasm-utils.cpp
rturrado: Very good catch! The join string in `cqasm-utils.hpp`, `to_json` was set to `""` instead of `","`.
That is probably good JSON because it is probably interpreted as a single string, but it should in fact be interpreted as a vector of strings.

include/v3x/cqasm-analyzer.hpp
rturrado: Thanks!

python/module/cqasm/v3x/__init__.py
rturrado: I have added some tests, and I don't have here the same problem as I've had in other PR, that the tests are not found, they are found, but then there are errors executing those tests. I think it has to do with SWIG, and modules, and making those modules accessible from Python code.

python/module/libQasm/__init__.py
```diff
@@ -1,17 +1,21 @@
 # Author Imran Ashraf, Jeroen van Straten
 
-# Before we can import the dynamic modules, we have to set the linker search
-# path appropriately.
 import os
+
+from sys import version_info
```
pablolh, line 5: double import?

rturrado: Fixed!

pablolh:
Sorry for the delay!

rturrado:
> Sorry for the delay!

No worries Pablo. I've merged `develop` into `json_dumper`, fixed some things I screwed up during the merge, and now I need your approval again.

pablolh (approved):

PR #168, opened 2023-11-24
One single version of truth

rturrado:
Added `include/version.hpp` to define `libqasm` version in a single place.
Then made this definition available to Python code through the `get_version` method at `version.py`.
And used this method from `setup.py`.

This is very similar to what is done at the moment for the QX simulator.
The main difference is that I've put the implementation of the get_version method in a `version.py` file.
I've done this because I think that, in the future, it may be needed to access this `get_version` method from `conanfile.py`.

### DIFF
--- a/version.py
+++ b/version.py
@@ -1,0 +1,22 @@
+import os
+import re
+
+
+def get_version(verbose=0):
+    """Extract version information from source code"""
+
+    root_dir = os.getcwd()  # root of the repository
+    inc_dir = root_dir + os.sep + "include"  # C++ include directory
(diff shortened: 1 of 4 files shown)


version.py
```diff
@@ -0,0 +1,22 @@
+import os
+import re
+
+
+def get_version(verbose=0):
```
pablolh, line 5: Python has booleans! `True` is better than `0`

rturrado: You mean `False`, right? Sorry about this, I just copy-pasted from the QX simulator project.

pablolh (approved):

PR #182, opened 2023-12-15
Release 0.5.0

rturrado:
A new release feels like changing the minor version. Last release dates from July 15, 2021. 😀 

Basically all the changes from this PR are focused on `assets.yml`, to make `manylinux-x64` and `manylinux-arm64` jobs work.

I have been forced to downgrade the gtest Conan package from `1.14` to `1.13`. I was getting a linkage error in the `manylinux-arm64` job, an undefined reference in `gmock-all.cc` to `__libc_single_threaded`. I read [here](https://groups.google.com/g/sparselizardforum/c/bfmP65SP9gE/m/Zg8KyuM2AgAJ) that this had to do with a recent version of gcc being needed. And it occured to me that, instead of trying to play with gcc versions, I may as well play with the gtest version.  And it worked.

I have also taken the opportunity to:
- do a couple of small changes in `src/CMakeLists.txt` (mainly removing a duplicated `find_package(fmt)`).
- add a `CHANGELOG.md`, mimicking what whe do in the QX simulator project.


.github/workflows/assets.yml
```diff
@@ -112,21 +128,37 @@ jobs:
...
         run: |
-          /opt/python/${{ matrix.cpython_version }}/bin/python -m pip install --upgrade pip conan wheel auditwheel
+          dnf install -y $BISON_VERSION $FLEX_VERSION $JAVA_VERSION $SWIG_VERSION
+          python_bin="/opt/python/${{ matrix.cpython_version }}/bin/python"
+          ${python_bin} -m pip install --upgrade pip conan wheel auditwheel
+          pip_installation_folder=$(${python_bin} -m pip show --files conan | grep 'Location:' | sed 's/Location: //')
+          conan_installation_folder_relative_path=$(${python_bin} -m pip show --files conan | grep '/bin/' | sed 's#\s\+\(.*\)/conan#\1#')
+          conan_installation_folder="${pip_installation_folder}/${conan_installation_folder_relative_path}"
+          echo "Conan installation folder: ${conan_installation_folder}"
+          echo "${conan_installation_folder}" >> $GITHUB_PATH
```
pablolh, line 152: Is it not enough to do
`export PATH="/opt/python/${{ matrix.cpython_version }}/bin:$PATH"`
like in QX?

pablolh: (or equivalently with GITHUB_PATH)

oschusler (approved):

PR #188, opened 2024-01-08
Fix GCC13 build by including <cstdint>

pablolh:
and add std:: prefixes


rturrado (approved):
Beautiful, Pablo, thanks for the fix!

PR #191, opened 2024-01-17
Remove msvc-runtime Python Windows dependency

pablolh:
This package doesn't seem well maintained and libqasm works fine on Windows without. Moreover, it's creating issues with Python 3.12 (impossible to install libqasm).


rturrado (approved):
Great! As long as anything breaks, all code removals are welcome! 😀

PR #194, opened 2024-01-22
V3 Extended: Arithmetic Expressions

rturrado:
Add arithmetic expressions.
These include unary, binary, and ternary operators, as well as a set of default functions (e.g. `sin`, `cos` and so on).


include/v1x/cqasm-analyzer.hpp
```diff
@@ -127,9 +126,9 @@ class Analyzer {
     );
 
     /**
-     * Convenience method for registering a function. The param_types are
```
pablolh, line 130: Do you use clang-format or the like?
Maybe we could run it on the whole codebase, instead of fixing here and there

src/v3x/AnalyzeTreeGenAstVisitor.cpp
```diff
@@ -384,23 +384,146 @@ std::any AnalyzeTreeGenAstVisitor::visit_expression(ast::Expression &node) {
...
-std::any AnalyzeTreeGenAstVisitor::visit_boolean_literal(ast::BooleanLiteral &node) {
-    auto ret = tree::make<values::ConstBool>(node.value);
-    return values::Value{ ret };
+/*
+ * Convenience function for visiting a function call given the function's name and arguments
+ */
+values::Value AnalyzeTreeGenAstVisitor::visit_function_call(
+    const tree::One<ast::Identifier> &name,
+    const tree::One<ast::ExpressionList> &arguments) {
+
```
pablolh, line 393: blank line after `{` feels unexpected

src/v3x/BuildTreeGenAstVisitor.cpp
```diff
@@ -321,32 +321,121 @@ std::any BuildTreeGenAstVisitor::visitIndexRange(CqasmParser::IndexRangeContext
...
+std::any BuildTreeGenAstVisitor::visitPowerExpression(CqasmParser::PowerExpressionContext *context) {
+    auto ret = tree::make<PowerExpression>(std::any_cast<One<Expression>>(context->expression()->accept(this)));
+    const auto token = context->POWER_OP()->getSymbol();
+    setNodeAnnotation(ret, token);
+    return One<Expression>{ ret };
+}
+
+std::any BuildTreeGenAstVisitor::visitProductExpression(CqasmParser::ProductExpressionContext *context) {
+    if (context->PRODUCT_OP()) {
+        return visitBinaryExpression<ProductExpression>(context, [context]() { return context->PRODUCT_OP(); });
```
pablolh, line 361: Something I learnt recently is that in C++ lambdas you do not need to write `()` when there is no argument

include/v3x/BuildTreeGenAstVisitor.hpp
```diff
@@ -37,6 +37,17 @@ class  BuildTreeGenAstVisitor : public BuildCustomAstVisitor {
     std::any visitFloatTypeDefinition(CqasmParser::FloatTypeDeclarationContext *context);
     std::any visitFloatTypeInitialization(CqasmParser::FloatTypeDeclarationContext *context);
 
+    template <typename RetExpressionType, typename Context, typename GetTerminalNodeFunction>
+    std::any visitBinaryExpression(Context *context, GetTerminalNodeFunction&& getTerminalNode) {
+        auto ret = tree::make<RetExpressionType>(
+            std::any_cast<tree::One<cqasm::v3x::ast::Expression>>(context->expression(0)->accept(this)),
+            std::any_cast<tree::One<cqasm::v3x::ast::Expression>>(context->expression(1)->accept(this))
+        );
+        const auto token = getTerminalNode()->getSymbol();
```
pablolh, line 46: I'm a bit confused here, so just for my information
What is the reason to pass getTerminalNode lambda and not directly the terminal node or the token?

src/v3x/BuildTreeGenAstVisitor.cpp
```diff
@@ -361,6 +450,13 @@ std::any BuildTreeGenAstVisitor::visitIndex(CqasmParser::IndexContext *context)
     return One<Expression>{ ret };
 }
 
+std::any BuildTreeGenAstVisitor::visitIdentifier(CqasmParser::IdentifierContext *context) {
+    auto ret = cqasm::tree::make<Identifier>(context->IDENTIFIER()->getText());
+    const auto token = context->IDENTIFIER()->getSymbol();
+    setNodeAnnotation(ret, token);
```
pablolh, line 456: what is the node annotation used for?

include/v1x/cqasm-analyzer.hpp
rturrado: That's a very good point. I do not use `clang-format`. It's in my TODO list of things to apply. I just love when CLion gives you all these PEP warnings in Python code, so something similar for C++ code would be great.

src/v3x/AnalyzeTreeGenAstVisitor.cpp
rturrado: You'll find some code in the codebase writing only `) {` in a line, and then the function body. I prefer to put the `) {` after the last parameter but, at the same time, I don't like the code starting with the same indentation as the last parameter so, for these cases, I leave a blank line there. But I can remove it or write only `) {` in a line.

include/v3x/BuildTreeGenAstVisitor.hpp
rturrado: Wow. There is no reason!

include/v1x/cqasm-analyzer.hpp
rturrado: But I understand that with `clang-format` you can just specify a line width, right? I try to reformat comments so that 1) every line is quite meaningful on its own, and 2) a sentence is not split in two lines. I find that eases the reading.

pablolh: I am indeed not sure if clang-format has option for this kind of formatting...

src/v3x/AnalyzeTreeGenAstVisitor.cpp
pablolh: OK, then, that's a choice, I'm OK with it

src/v3x/BuildTreeGenAstVisitor.cpp
pablolh: What about this?

pablolh (dismissed):

include/v1x/cqasm-functions.hpp
```diff
@@ -0,0 +1,232 @@
...
+
+/*
+ * Ternary function with constant parameters 
+ */
+template <typename ParamType, auto F>
+struct tf_cp : public f_cp<ParamType, ParamType, F> {
+    values::Value operator()(const values::Values &vs) const {
+        values::check_const(vs);
+        auto condition = vs[0]->as_const_bool()->value;
+        auto if_true = dynamic_cast<ParamType&>(*vs[1]).value;
```
pablolh, line 84: If we meet Jeroen for lunch at X again, he would tell you to use `vs[1].as<ParamType>()` I believe

include/v1x/cqasm-functions.hpp
```diff
@@ -0,0 +1,232 @@
...
+constexpr auto asin = [](auto num) { return std::asin(num); };
+constexpr auto fn_asin_c = uf_cp<values::ConstComplex, values::ConstComplex, asin>{};
+constexpr auto fn_asin_r = uf_cp<values::ConstReal, values::ConstReal, asin>{};
+constexpr auto acos = [](auto num) { return std::acos(num); };
+constexpr auto fn_acos_c = uf_cp<values::ConstComplex, values::ConstComplex, acos>{};
+constexpr auto fn_acos_r = uf_cp<values::ConstReal, values::ConstReal, acos>{};
+constexpr auto atan = [](auto num) { return std::atan(num); };
+constexpr auto fn_atan_c = uf_cp<values::ConstComplex, values::ConstComplex, atan>{};
+constexpr auto fn_atan_r = uf_cp<values::ConstReal, values::ConstReal, atan>{};
+constexpr auto asinh = [](auto num) { return std::asinh(num); };
```
pablolh, line 199: Can you not directly use `std::asinh` in `uf_cp` template parameters?
I never tried myself

test/v3x/cpp/AnalyzeTreeGenAstVisitor.cpp
```diff
@@ -0,0 +1,66 @@
...
+    auto analyzer = analyzer::MockAnalyzer();
+    EXPECT_CALL(analyzer, call_function(::testing::_, ::testing::_))
+        .WillOnce(::testing::Return(function_return_value));
+    auto visitor = analyzer::AnalyzeTreeGenAstVisitor(analyzer);
+
+    auto name = cqasm::tree::make<ast::Identifier>(function_name);
+    auto arguments = cqasm::tree::make<ast::ExpressionList>(cqasm::tree::Any<ast::Expression>{});
+    auto function_call = ast::FunctionCall{ name, arguments };
+    auto ret = std::any_cast<values::Value>(visitor.visit_function_call(function_call));
+    EXPECT_EQ(ret->as_const_real()->value, function_return_value->as_const_real()->value);
```
pablolh, line 32: Minor question
Can you use tree-gen's `.equals` here? It is made (and built-in) for comparing tree-gen trees

test/v3x/cpp/AnalyzeTreeGenAstVisitor.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+
+TEST(visit_function_call, analyzer_call_function_returns_an_empty_value) {
```
pablolh, line 35: I really like that there are now proper unit tests :)

To make them more readable, now or in the future, we can use TEST_FIXTURE's instead, and a test class. This test class could contain common test data, so that you can write shorter tests where you instantly know what's going on, because they're short.

For instance, your test base class could already declare the MockAnalyzer as a member, then each test can use it without declaring it.

Another idea is to declare some aux functions like `expect_call_function` or the like, since you use it at least in 2 places.

For now it's not required, but should we add more unit tests, fixtures are really nice for factoring test code.

test/v3x/cpp/matcher_values.cpp
```diff
@@ -0,0 +1,36 @@
...
+namespace cqasm::v3x::values {
+
+ValuesEqMatcher::ValuesEqMatcher(const Values &expectedValue)
+: expectedValue_(expectedValue)
+{}
+
+bool ValuesEqMatcher::MatchAndExplain(const Values &args, std::ostream* /* os */) const {
+    const auto &expected_vec = expectedValue_.get_vec();
+    const auto &actual_vec = args.get_vec();
+    return std::equal(expected_vec.begin(), expected_vec.end(), actual_vec.begin(), actual_vec.end(),
```
pablolh, line 18: You can replace this `std::equal` with what `tree::Any` already has `.equals()`, see https://github.com/QuTech-Delft/tree-gen/blob/e365950b56752c6de53025b002a96ed905e9ddae/include/tree-base.hpp.inc#L993

pablolh (approved):
no problem for me, just nitpicking

test/v3x/cpp/AnalyzeTreeGenAstVisitor.cpp
rturrado: Something like `EXPECT_TRUE(ret.equals(function_return_value));`? I'll give it a go, thanks!

test/v3x/cpp/AnalyzeTreeGenAstVisitor.cpp
rturrado: I'll take that into account, yes, thanks!

test/v3x/cpp/matcher_values.cpp
rturrado: I'll have a look, thanks!

test/v3x/cpp/AnalyzeTreeGenAstVisitor.cpp
pablolh: yes I think so
It also checks if nodes have the same types so it should work I believe

include/v1x/cqasm-functions.hpp
rturrado: I've tried this again. But, using `auto F` as a non-type template parameter, we can pass a functor, e.g., `std::negate{}`, or a lambda, e.g., our `asinh`.

We could pass `std::asinh` directly by using a combination of template parameters and a non-type template parameter, e.g., `typename FReturnType, typename FParamType, FReturnType (*F)(FParamType)`, but that would not go well along with passing a functior.

test/v3x/cpp/matcher_values.cpp
rturrado: Done!

test/v3x/cpp/AnalyzeTreeGenAstVisitor.cpp
rturrado: Done!

test/v3x/cpp/AnalyzeTreeGenAstVisitor.cpp
rturrado: Done!

pablolh (approved):

PR #199, opened 2024-03-07
Enable libqasm for emscripten

rturrado:
libqasm can now be cross-compiled to emscripten.

This cross-compilation is tested via GitHub Actions from an ubuntu/x64 machine.

### `.github` folder
- `actions/cpp-tests/action.yaml`: use the same profile both for build and for host.
- `workflows/test.yml`: add `js-emscripten-wasm` job.

### `emscripten` folder
- `CMakeLists.txt`: CMake code to generate the emscripten binaries, `cqasm_emscripten.js` and `cqasm_emscripten.wasm`.
- `emscripten_wrapper.hpp`: define the bindings that are exported to Javascript callers: `get_version`, `parse_string_to_json`, and `analyze_string_to_json`.
- `emscripten_wrapper.cpp`: connect the bindings to their implementations, `V3xAnalyzer` methods.
- `test_cqasm.js`: exercise the bindings via different test cases:  get version, parse returning an error or an AST, and analyze returning an error or an AST.

### `src` folder
- `CMakeLists.txt`: when building for emscripten, we check out and build ANTLR, `fmt`, and `tree-gen` code, so that exceptions risen in this code can be correctly treated in the generated wasm code.

### Root folder
- `CMakeLists.txt`: for emscripten binaries, we have to enable exception support.
- `conanfile.py`: update to use tree-gen Conan package.
- `Dockerfile`: install latest emsdk, install CMake 3.28, invoke `node` from emsdk.
- `README.md`: update.
- `setup.py`: use the same profile both for build and for host.


emscripten/emscripten_wrapper.cpp
```diff
@@ -0,0 +1,47 @@
+#include "emscripten_wrapper.hpp"
+
+#include "v3x/cqasm-py.hpp"
+#include "version.hpp"
+
+
+EmscriptenWrapper::EmscriptenWrapper() {}
```
pablolh, line 7: you could just `= default` this in the header

emscripten/test_cqasm.js
```diff
@@ -0,0 +1,77 @@
...
+
+process.on("unhandledRejection", error => {
+    console.error("unhandledRejection", error, "\n");
+});
+
+wrapper().then(function(result) {
+    var cqasm = new result["EmscriptenWrapper"]()
+    var ret_code = 0
+
+    console.log("Hola from JavaScript!")
```
pablolh, line 11: Hola!

emscripten/test_cqasm.js
```diff
@@ -0,0 +1,77 @@
...
+
+    try {
+        var output = cqasm.get_version()
+        var expected_output = "0.5.2"
+        console.log("\nThe version of cqasm compiled with emscripten is:", output);
```
pablolh, line 16: is it the version of cqasm or the version of libqasm?

pablolh (approved):

emscripten/emscripten_wrapper.cpp
rturrado: Thanks, good spot!

emscripten/test_cqasm.js
rturrado: `libqasm`. I'll update the text!

pablolh (approved):

PR #212, opened 2024-04-08
Fix Python primitives

rturrado:
`python/module/cqasm/v3x/primitives.py` still contains references to `Real` type, which is now called `Float`.


elenbaasc (approved):

elenbaasc (commented):
This seems to be in order :P

PR #216, opened 2024-04-12
Create master beta 3

rturrado:
### Added

- `cpp-linux-x64-release-shared` test job.

### Changed

- Allow multiple `measure` instructions at the end of a program.

### Removed


elenbaasc:
Do we want to add more features to this 'version' before we release?

rturrado:
> Do we want to add more features to this 'version' before we release?

I don't think so, Chris. I would take this as it is a a release candidate, and go on a code freeze (only bug fixes allowed).

elenbaasc (approved):

PR #223, opened 2024-04-26
Allow multiple qubit/bit (register) definitions and mid-circuit measurements

rturrado:
A program must have a version, then can optionally have statements.
Statements can be variable definitions or instructions.
Instructions can be gates or measure instructions.

Also:
generate_antlr_parser.py: fix shutil.move when destination file exists.


elenbaasc (approved):
All looks good.

The `res/v3x/parsing` contains all these test inputs and golden files. Is it necessary to do the testing this way? Maybe not something to refactor now, but wouldn't it be better to not have this being file-based?

rturrado:
> All looks good.
> 
> The `res/v3x/parsing` contains all these test inputs and golden files. Is it necessary to do the testing this way? Maybe not something to refactor now, but wouldn't it be better to not have this being file-based?

Thanks, @elenbaasc.

Yes, indeed.

Basically, the idea is to move away, not completely, but limit those file-based integration tests to a minimum, and do more unit-testing.

But, since I already had those tests, I'm basically recovering them from the backup and putting them back on again.

PR #236, opened 2024-07-26
[CQT-27] Fix MacOS/x64 (with Python >= 3.11) wheels

rturrado:
Issue:
We are generating universal instead of x86_64 binaries.
According to this [thread](https://github.com/pypa/wheel/issues/573), it's highly recommended to move to cibuildwheel as wheel creator.

Changes:
- Use cibuildwheel for generating the wheels.
- Remove support for Python 3.8 (the combo cibuildwheel/macos14/CPython 3.8 didn't work).
- Add pyproject.toml.
- Update setup.py: remove bdist, bdist_wheel, sdist, and egg_info.

### DIFF
--- a/.github/workflows/assets.yml
+++ b/.github/workflows/assets.yml
@@ -9,247 +9,61 @@ on:
       - "release**"
 
 jobs:
-  macos-x64:
-    name: PyPI wheels for macOS/x64
-    runs-on: macos-13  # x64
+  cibw-wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python:
-          - "3.8"
-          - "3.9"
-          - "3.10"
-          - "3.11"
-          - "3.12"
+        # macos-13 is an x64 runner, macos-14 is an arm64 runner
+        os: [ubuntu-latest, windows-latest, macos-13, macos-14]
     steps:
       - name: Checkout
         uses: actions/checkout@v4
       - uses: actions/setup-python@v5
-        with:
-          python-version: ${{ matrix.python }}
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip conan setuptools wheel
-          brew install swig
-      - name: Build wheel
-        run: python setup.py bdist_wheel
-      - name: Wheel path
-        id: wheel
-        working-directory: pybuild/dist/
-        run: |
-          echo "WHEEL_NAME=$(ls *.whl)" >> $GITHUB_OUTPUT
-          echo "WHEEL_NAME=$(ls *.whl)" >> $GITHUB_ENV
+      - name: Install cibuildwheel
+        run: python -m pip install cibuildwheel==2.18.1
+      - name: Build wheels
+        run: python -m cibuildwheel --output-dir wheelhouse
       - uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -12,19 +12,31 @@
...
 from distutils.command.build import build as _build
 from setuptools.command.install import install as _install
-from distutils.command.bdist import bdist as _bdist
-from wheel.bdist_wheel import bdist_wheel as _bdist_wheel
-from distutils.command.sdist import sdist as _sdist
-from setuptools.command.egg_info import egg_info as _egg_info
 
-from version import get_version
+
+# TODO: I had to copy-paste get_version from version.py here
```
oschusler, line 17: The code should not have todo's. If there is still work to be done, this should be in a ticket in the ticketing system

setup.py
```diff
@@ -12,19 +12,31 @@
...
+#  because I couldn't get 'from version import get_version' work with pyproject.toml
```
oschusler, line 18: What's with the pyproject.toml?

.github/workflows/assets.yml
```diff
@@ -307,7 +117,8 @@ jobs:
         id: download
       - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@v1.8.14
+        if: ${{ github.event_name == 'release' && github.event.action == 'created' }}
```
oschusler, line 120: Where is the source code for deploying to TestPyPI?

oschusler (changes requested):

setup.py
rturrado: Good point. Thanks!

rturrado: I've removed the TODO and created issue #240.

setup.py
rturrado: I've explained this in issue #240.

.github/workflows/assets.yml
rturrado: I've removed it. Copy-pasting here a comment I've just made in Jira.

> I’m sorry for the misundersandings and the confusion.
> 
> In my mail (24 Jul 2024, 16:13) I suggested dropping the TestPyPI step because, first, it was not behaving as I had expected, and second, it was giving me problems to set it up.
> 
> I didn’t understand that your response mail (25 Jul 2024, 21:34) was a negative to drop the TestPyPI step, but just an explanation of what you considered the value of using TestPyPI was. This is the complete contents of that mail, except for greetings:
> 
> > The goal however from deploying the packages to TestPyPI, never was about being able to test the released packages. For me, it is about testing whether the release actually succeeds.
> 
> There were two other facts that made me drop this step:
> 
> It wasn’t something required for this ticket. This ticket intends to fix a given libqasm build. But, since I planned to suggest creating a release after this ticket, I thought it would be good to have that publishing to TestPyPI in.
> 
> None of the cQASM tooling projects do a publishing to TestPyPI.
> 
> So, things to learn from this:
> 
> I should take more care deviating from the exact intents of each ticket. This can cause extra work, delays (not really lost work in this case, but it could be in other cases)…
> 
> I wouldn’t really add a TestPyPI step (maybe just because I still don’t understand neither its benefits, nor how to overcome its issues) but, if we want it, we can use another ticket for it.

setup.py
oschusler: Tickets should be moved to Jira

PR #244, opened 2024-08-19
[CQT-6] Parsing and analyzing the same file with a syntactic error can return different messages

rturrado:
When libqasm was supporting both v1 and v3 files parsing went to the v1 or v3 parsers directly, but analyzing went to a version parser first stage.
Depending on the version, libqasm would route the analysis to the v1 (Flex/Bison) or the v3 (ANTLR) analyzers.
Syntactic errors such as the program 'vrsion 3' would produce an early error from the version parser (also Flex/Bison).
Nowadays libqasm only supports v3, and all the parsing and analyzing goes directly to ANTLR.
Thus, for programs such as 'vrsion 3', both parsing and analysing report the same error.

- Add a Python test to verify this with a simple example.
- Add an integration test as well to cover the program 'vrsion 3'.
  This program case was covered before by the version parser path.
  But now that we don't have the version parser anymore, it's good to add it as an integration test.


elenbaasc (approved):
Looks good 👍

PR #245, opened 2024-08-21
[CQT-125] Automate doc publishing

rturrado:
Add a  `doc` workflow that builds `libqasm` docs and, when needed, publishes them to GitHub pages.
The new workflow is based on cQASM-spec `release` workflow.

- Remove Conan and CMake options for building the docs (use `mkdocs build` instead).
- Remove custom `doc/mkdocs`.
- Update `doc.yml` workflow to work with `mike`.
- Mimic the documentation framework from cQASM-spec:
  - Rename `doc` folder to `docs`.
  - Place `mkdocs.yml` and `mkdocs-base.yml` at the root folder.
  - `mkdocs-base.yml` references to files in `docs` do not include any prefix (cQASM-spec uses a `./` prefix)
  - The styles should be very similar to cQASM-spec. The only difference, apart from light and white modes, should be that libqasm uses a css to correctly display the different APIs.


.github/workflows/doc.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Install dependencies
+        run: |
+          sudo apt-get install -y doxygen
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Install Python dependencies
+        run: |
+          python -m pip install --upgrade pip mike mkdocs-material mkdocstrings pymdown-extensions
+      - name: Build
```
oschusler, line 33: The job name is "Publish GitHub Pages", but there is an include of cpp-tests here, why?

.github/workflows/doc.yml
```diff
@@ -0,0 +1,60 @@
...
+        uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          sudo apt-get install -y doxygen
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Install Python dependencies
+        run: |
+          python -m pip install --upgrade pip mike mkdocs-material mkdocstrings pymdown-extensions
```
oschusler, line 32: This should be moved to a `requirements.txt` file. Dependabot will not pick up on the dependencies if we pin them, and they should be pinned. At the moment, we have no clue which version of mike (or other dependencies) are being used.

.github/workflows/doc.yml
```diff
@@ -0,0 +1,60 @@
...
+      - master
+  release:
+    types:
+      - created
+
+jobs:
+  publish-github-pages:
+    name: Publish GitHub pages
+    runs-on: ubuntu-latest
+    strategy:
```
oschusler, line 16: why do you have this strategy here? There is only one type

oschusler (changes requested):

.github/workflows/doc.yml
rturrado: Good point. I'm just reusing the C++ custom action that you added back in the day. And it turns out that that action runs tests as well. What do you reckon? Should we:
- Just don't use that action and duplicate the `pipx install conan` plus the libqasm build here?
- Rename the C++ custom action to `cpp-bulid`  receiving an input for optionally running the tests?
- ...

.github/workflows/doc.yml
rturrado: OK. Fine with me to pin the versions of the Python dependecies. But I don't think we are using Dependabot for the C++ projects.

.github/workflows/doc.yml
rturrado: Good point. I'm just reusing the C++ custom action that you added back in the day. And it turns out that that action runs tests as well. What do you reckon? Should we:
- Just don't use that action and duplicate the `pipx install conan` plus the libqasm build here?
- Rename the C++ custom action to `cpp-bulid`  receiving an input for optionally running the tests?
- ...

oschusler: I think we should revisit the splitting of the tests at some point, but or now duplicate the install of conan. Do we really need to build LibQASM btw for documentation deployment?

.github/workflows/doc.yml
oschusler: Dependabot doesn't support C++, so we don't have to turn it on for that. However, Docker images, python package, github actions, etc can be maintained by Dependabot so it would still be a nice addition.

.github/workflows/doc.yml
oschusler: There are only two places within that task that need the build type, right?

rturrado: Yes, yes, it can be hardcoded without problem.
But I would still leave `strategy` to include  the `fail-fast: false`.

oschusler: According to the documentation of GitHub actions:

> `jobs.<job_id>.strategy.fail-fast` applies to the entire matrix. If `jobs.<job_id>.strategy.fail-fast` is set to `true` or its expression evaluates to `true`, GitHub will cancel all in-progress and queued jobs in the matrix if any job in the matrix fails. This property defaults to true.

This means that only if you have multiple jobs in your matrix (running in parallel) the `fail-fast` property is relevant. They prevent your jobs continuing if one of the variants has already failed. Given that we only have one variant in the matrix, `fail-fast` has no meaning.

.github/workflows/doc.yml
rturrado: We build  libqasm with a conan _docs_ profile that basically adds a CMake target to invoke `mkdocs`.
But you're right that this shouldn't fire a complete build (i.e., we do a full build plus docs, whereas we should be doing just a docs build).
I'll have a look into this.

.github/workflows/doc.yml
rturrado: Ah OK, fair enough!

rturrado: Done!

.github/workflows/doc.yml
rturrado: Done!

rturrado:
Hi @oschusler, I think I have addressed your code review comments.

I have to admit that I have my concerns about a couple of things in this PR, namely:

- In order to build the documentation we can just go to `libqasm/doc` folder and run `python mkdocs build`, or `python mkdocs serve` (in case you also want to serve it).  
We could also, however, run Conan with a given Conan profile, and pass an option to build the docs, e.g., `conan build -pr:a=conan/profiles/release -o libqasm/*:build_docs=True -b missing`. The only benefit of using this option is to be able to install the document files during an installation. I'm not really sure if we want to get rid of this (`build_docs` Conan option, `doc/CMakeLists.txt` file, all the CMake code using `LIBQASM_BUILD_DOCS`...).

- Because we use our custom MkDocs handlers, we have our own `doc/mkdocs` Python script that wraps `mkdocs` and `mike deploy`. However, the `doc` GitHub workflow also invokes `mike set-default`. I am not 100% sure `mike set-default` is going to work fine after doing `python mkdocs deploy` (I would say so, because `mike set-default` shouldn't need anything from neither from `doc/mkdocs` nor from the custom MkDocs handlers.

oschusler (approved):

oschusler (approved):

PR #253, opened 2024-10-02
[CQT-191] [Infra] Fix GitHub pages Python syntax highlighting 


rturrado:
There is really nothing wrong with the Python syntax highlighting. It just is different from that of C++. For example, C++ syntax highlighting emphasizes more the function names while Python’s doesn't.

I’ve taken the opportunity to do a couple of changes though:
- Fix Emscripten syntax highlighting. Emscripten docs use C++ classes and functions (because Emscripten API is defined in a C++ wrapper). We were using Emscripten syntax highlighting, and I’ve changed it to C++.
- Change dark mode them from `zenburn` to `atom-one-dark`. I was trying a few, and I found this one the better for all C++, Emscripten and Python.
- Some minor aesthetic changes: unify the `__init__.py` custom handlers files, add a `:` at the end of the Python function declarations, and remove `;` at the end of Python code examples.


juanboschero (approved):
Looks good Roberto!

PR #256, opened 2024-10-10
[CQT-189] [Formatting] Add clang-format

rturrado:
Add cpp-linters job.
- At the moment, this job only executes clang-format.

Update .clang-format to use with LLV18:
- Add Standard and set it to c++20.
- Change LambdaBodyIndentation to Signature.
- Change PointerAlignment and ReferenceAlignment to Left.
  This change makes pointers and references to be written close to types.
  The counter indication to this would be `int* p, q;`.
  But that is a corner case and goes against the core guidelines of declaring multiple variables in the same line (especially if they are of different types).

Add scripts/run_cpp_linters.py.
- To allow running the C++ linters by hand through a Python script.

Update docs/dev-guide.md.


scripts/run_cpp_linters.py
```diff
@@ -0,0 +1,61 @@
...
+        file_list_str = " ".join(file_list)
+        command = f"clang-format-18 -i --style=file:{format_file_path} --verbose {file_list_str}"
+        subprocess.run(command, shell=True)
+    except FileNotFoundError as err:
+        print("Error running clang-format: {}".format(err.strerror))
+        exit(3)
+
+
+def run_clang_tidy(root_folder: os.PathLike):
+    pass
```
elenbaasc, line 45: I assume this is 'to be implemented', maybe a comment should be added that this is to be done.

docs/dev-guide/dev-guide.md
```diff
@@ -113,6 +127,23 @@ mkdocs serve
 
     The `export` is needed to point `mkdocs` to the custom handlers used for the C++, emscripten, and Python APIs.
 
+## Linters
+
+Continuous Integration will fail if the files don't adhere to a series of formatting and code style guidelines:
```
elenbaasc, line 132:
```suggestion
Continuous Integration will fail if the files do not adhere to a series of formatting and code style guidelines:
```

docs/dev-guide/dev-guide.md
```diff
@@ -47,6 +47,20 @@ For the time being, we install Java manually for this platform.
     - `mkdocstrings`
     - `pymdown-extensions`
 
+### Linters
+
+- `clang-format-18`
+- `clang-tidy-18`
+
+In a Linux machine, these linters can be installed with the following commands:
```
elenbaasc, line 55:
```suggestion
On a Linux machine, these linters can be installed with the following commands:
```

docs/dev-guide/dev-guide.md
```diff
@@ -86,7 +100,7 @@ All the `tests`, except for `linux-x64` profiles, enable Address Sanitizer.
 
 ### Options
 
-Profiles are a shorthand for command line options. The command above could be written, similarly,  as:
+Profiles are a shorthand for command line options. The command above could be written, similarly, as:
```
elenbaasc, line 103:
```suggestion
Profiles are a shorthand for command line options.
The command above could be written, similarly, as:
```

docs/dev-guide/dev-guide.md
```diff
@@ -113,6 +127,23 @@ mkdocs serve
...
+
+It is recommended to run these linters before pushing any change:
+
+```shell
+python3 ./scripts/run_cpp_linters.py .
+```
+
+!!! note
+
+    The linters require `clang-format-18` and `clang-tidy-18` to be installed in the system.
```
elenbaasc, line 145:
```suggestion
    The linters require `clang-format-18` and `clang-tidy-18` to be installed on the system.
```

.github/workflows/test.yaml
```diff
@@ -3,13 +3,39 @@ name: Test
 on:
   push:
     branches:
-    - develop
-    - master
+      - develop
+      - master
+      - CQT-189-based-off-make-libqasm-names-consistent
```
elenbaasc, line 8: This can be removed, right? Wouldn't want this on develop.

.github/workflows/test.yaml
```diff
@@ -3,13 +3,39 @@ name: Test
...
+          extensions: 'cpp,hpp'
+          files-changed-only: 'false'
+          format-review: 'true'
+          style: 'file'  # use .clang-format config file
+          tidy-checks: '-*'  # disable clang-tidy checks
+          version: 18
+      - name: Fail fast
+        if: steps.linter.outputs.clang-format-checks-failed > 0
+        run: |
+          echo "::notice::Try executing 'python3 ./scripts/run_cpp_linters.py .' to fix linter issues."
```
elenbaasc, line 33: To 'fix linter issues' or to 'display linter issues'?
These linters don't fix things, right? They just check and output linting errors. It is up to the dev to fix these.

.github/workflows/test.yaml
```diff
@@ -3,13 +3,39 @@ name: Test
...
+          exit 1
+
   cpp-linux-x64:
     name: "C++ tests (gcc-clang/Linux/x64)"
+    needs: cpp-linters
```
elenbaasc, line 38: Do these linters need to run for each architecture? If one is fine, all should be fine right?

.clang-format
```diff
@@ -17,15 +17,22 @@ AllowShortLoopsOnASingleLine: true
 AlwaysBreakTemplateDeclarations: Yes
 BinPackArguments: false
 BreakBeforeBraces: Attach
+# Use BraceWrapping with custom BreakBeforeBraces
+# BreakBeforeBraces: Custom
+# BraceWrapping:
```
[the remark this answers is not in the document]
rturrado: I can. But more than commented out lines, they are an explanation: "if you ever need to be more precise regarding braces, know that you can set BreakBeforeBraces to Custom and then set BraceWrapping to whatever".

scripts/run_cpp_linters.py
```diff
@@ -0,0 +1,61 @@
...
+import sys
+
+
+def print_usage():
+    print("Usage:")
+    print("    python3 run_cpp_linters.py <ROOT_FOLDER>")
+    print("Where:")
+    print("    ROOT_FOLDER: folder containing the .clang-format, .clang-tidy, and C++ files.")
+    print("Example:")
+    print("    python3 run_cpp_linters.py .")
```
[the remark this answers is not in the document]
rturrado: I always do this with command line scripts. It's a help in case you don't know how to run the command. It's not really a logging mechanism.

scripts/run_cpp_linters.py
```diff
@@ -0,0 +1,61 @@
...
+        for root, dirs, files in os.walk(input_folder):
+            for file in files:
+                for file_type in file_types:
+                    if file.endswith(file_type):
+                        ret += [str(os.path.join(root, file))]
+    return ret
+
+
+def run_clang_format(root_folder: os.PathLike):
+    print("Running clang-format")
```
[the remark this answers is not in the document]
rturrado: As I  said before, I find `logging` a bit too (can't find the word... overblown?) for these type of scripts.

scripts/run_cpp_linters.py
rturrado: I planned to do that next week, and I didn't really want to put a TODO comment here.

.github/workflows/test.yaml
rturrado: Yes, thanks!

.github/workflows/test.yaml
rturrado: Yes, but if you run `python3 ./scripts/run_cpp_linters.py .` they will fix the issues for you. Linters have a _dry-run_ mode as well where they just display issues without fixing them. Think of them as `ruff` for OpenSquirrel.

.github/workflows/test.yaml
rturrado: Good point. I would say so. They just look at the source files, not at the built binaries.

docs/dev-guide/dev-guide.md
rturrado: These two sentences fit perfectly on a line. So discarding this suggestion.

rturrado:
> Nice work!

Many thanks for the review, Chris.

> Does PointerAlignment and ReferenceAlignment to Left, follow any particular conventions? I though the default for pointer was 'right' and reference was 'left'.

This is one of those eternal discussions within the C++ community. I personally prefer to write the pointer and the reference close to the type, because they are actually type modifiers. So when you read `int* p` or `int& r` you know `p` is a variable of type `int*` (pointer to integer), whereas `r` is a variable of type `int&` (reference to integer). Right-hand side advocates always say that `int* p, q` is confusing, because you may think that `q` is a pointer to integer while it really is an integer. While this is true, it is also true that `int* p, q` goes against the C++ coding guidelines, since you shouldn't declare more than one variable per line, especially when they are of different types. Apart from the fact that that is a corner case.

> Mostly minor comments, and in the python module for the linting, I think we should use `logging` iso the print statements.

My point against going beyond `print` is that these are really simple helper scripts, kind of tiny command line tools. I understand that, if you wanted to do a more serious script/app, `logging` would be the way to go, also having the possibility to write to standard output, standard error, file... and configure it. But I don't really see it here.

.clang-format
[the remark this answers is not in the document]
elenbaasc: As you wish

scripts/run_cpp_linters.py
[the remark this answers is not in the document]
elenbaasc: Then for this I guess we can use a print statement. Maybe nicer as one print call.
```suggestion
def print_usage():
    print("Usage:\n"
          "\tpython3 run_cpp_linters.py <root-folder>\n"
          "Where:\n"
          "\troot-folder: folder containing the .clang-format, .clang-tidy, and C++ files.\n"
          "Example:\n"
          "\tpython3 run_cpp_linters.py .")
```

scripts/run_cpp_linters.py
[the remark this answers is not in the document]
elenbaasc: Sure we leave as is, unless you feel inclined to try out logging (it's quite straightforward, but I leave it to you)

docs/dev-guide/dev-guide.md
elenbaasc: Hmmm, this goes against what you wanted for cQASM spec right? You wanted to at least have each separate sentence on a new line. Or did I misunderstand...

.github/workflows/test.yaml
elenbaasc: Ah, oke, so not just checking but also fixing. Good to know

elenbaasc:
Thanks for the explanation on the C++ conventions, make sense to attach to the type!

Still some small comments. Let me know when you're ready for me to have another look!

scripts/run_cpp_linters.py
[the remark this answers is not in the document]
rturrado: OK, I've used in the end a literal string.

scripts/run_cpp_linters.py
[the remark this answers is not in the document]
rturrado: OK, I've used it in the end. It made more sense for the other script, for `generate_antlr_parser.py`.

BTW, I know that script is not part of this PR, but I consider it justified to update it so both scripts look alike.

elenbaasc (approved):
All good!

rturrado:
You won't believe me, but this is one of the PRs that makes me more illusion of all.

I don't have to worry anymore about code formatting! 😅👯

PR #257, opened 2024-10-10
[CQT-216] Update image for ARM64 instances

oschusler:


rturrado (approved):
Looks fine to me, as long as, of course:
- it continues to pass tests, and
- we don't have a regression in performance for gcc/Linux/ARM64.

PR #259, opened 2024-10-16
Fix instruction set types

rturrado:
CZ and CRk types were interchanged.


elenbaasc (approved):
Perfect! Thanks for incorporating the fix!

PR #261, opened 2024-10-16
Run cpp-linux-arm64 job only when merging to `develop`

rturrado:
GitHub Actions takes about 12 minutes without this job, and 1.5 hours with it.
We can limit this job for merges to `develop` and still play safe.
GitHub has announced they will provide GitHub-hosted runners for Linux/ARM64 by the end of the year.
We'll make sure of enabling this job back for any branch then.


oschusler (approved):

elenbaasc (approved):

oschusler (changes requested):
Shouldn't this be develop?

oschusler (approved):

PR #263, opened 2024-10-17
Update `master` to `release-0.6.8`

rturrado:
Sync `master` from current `develop`.

---

I've found an issue with the `cpp-linter-action`. When it communicates with GitHub Actions, it receives a list of changes, and when this list is too big (I think the threshold is over 2'000 lines), the action fails. I've already reported this [issue](https://github.com/cpp-linter/cpp-linter-action/issues/274).

For the time being, what I've done is:
1. Take the C++ linters job out from the Test workflow and into a Linters workflow.
2. Remove the C++ linters job from the Test workflow.
3. Run the C++ linters job only when merging to `develop` or pulling request to `develop`.

This shouldn't be a problem because the C++ linters are tools that just perform some static checks (formatting at the moment via `clang-format`, and code style in a future via `clang-tidy`) on the codebase. And, with PRs like this, we are just syncing the codebase from `develop` to `master`.

However, there is yet an issue remaining: we could perfectly be doing a merge to `develop` or a pull request to `develop` making this `cpp-linter-action` to fail. In order to solve that in the future:
1. Either we have a fix for the issue mentioned above.
2. Or we don't make the `cpp-linters` job fail the workflow.


elenbaasc:
> Sync masterfrom current develop.
> 
> I've found an issue with the `cpp-linter-action`. When it communicates with GitHub Actions, it receives a list of changes, and when this list is too big (I think the threshold is over 2'000 lines), the action fails. I've already reported this [issue](https://github.com/cpp-linter/cpp-linter-action/issues/274).
> 
> For the time being, what I've done is:
> 
> 1. Take the C++ linters job out from the Test workflow and into a Linters workflow.
> 2. Remove the C++ linters job from the Test workflow.
> 3. Run the C++ linters job only when merging to `develop` or pulling request to `develop`.
> 
> This shouldn't be a problem because the C++ linters are tools that just performs some static checks (formatting at the moment via `clang-format`, and code style in a future via `clang-tidy`) on the codebase. And, with PRs like this, we are just syncing the codebase from `develop` to `master`.
> 
> However, there is yet an issue remaining: we could perfectly be doing a merge to `develop` or a pull request to `develop` making this `cpp-linter-action` to fail. In order to solve that in the future:
> 
> 1. Either we have a fix for the issue mentioned above.
> 2. Or we don't make the `cpp-linters` job fail the workflow.

I guess a more frequent release cycle might also help here (i.e. less changes in between the separate releases). It is strange though that there is this cap on the amount of changes... 

I guess it should be oke to take out the linting checks when merging to master (as long as they are there when merging to develop), do you agree @oschusler? If we get into issues when merging to develop we will need to come up with some other solution (if we haven't found one already)

2bndy5:
> It is strange though that there is this cap on the amount of changes...

It actually isn't strange at all when you think about the technology used (REST API). All git servers have a limit as to what their REST API can aggregate and return. GitLab's limit is something like 15Kb or 1000 lines (and they don't support getting the full diff in 1 API endpoint like GitHub and Gitea). Gitea and BitBucket also have similar limits, but I don't know their specific limits off the top of my head.


<details><summary>What about <code>git diff</code>?</summary>
<p>

[`actions/checkout`](https://github.com/actions/checkout) uses shallow checkouts by default, so using `git diff` just returns the entire history of the repo in that scenario. My experiments with extending the `fetch` depth yielded other problems related to permissions of the `github.token`. 😞 

</p>
</details> 

---
If all contributions to a development branch are checked by linters, then (IMHO) it seems redundant to run the linters again when merging known-good sources from a development branch to a release branch.

rturrado:
Many thanks for the detailed comment,  @2bndy5 !

oschusler:
I agree with @2bndy5 that linting should no longer have to happen after feature branches have been merged. However, it does depend on pinning of dependencies. We should investigate and discuss this at some point, but for now, lets take the checks out.

rturrado:
Sure, the checks are out since yesterday.

elenbaasc (approved):

PR #266, opened 2024-11-12
[CQT-253] Add instructions and directives to libQASM

rturrado:
- Add `SWAP` (unitary instruction).
- Add `init`, `barrier`, and `wait` (non-unitary instructions).
- Change `reset` instruction to always need qubit operands.


elenbaasc:
Thanks for the work; looks great!

I see some changes to the semantic analyzer, are these refactors to handle parameters better?

Any idea why the (emscripton/python) tests are failing?

rturrado:
> Thanks for the work; looks great!

Thanks!

> I see some changes to the semantic analyzer, are these refactors to handle parameters better?

Yes. When you add new code, you are constantly looking at improving existing code. In this case, we had a `resolve_gate` function who was only doing some semantic checks on the parameter type for a gate. With the introduction of parametrized non-instruction gates (`wait`), we reuse that function, changing its name to `resolve_parameter`.
 
> Any idea why the (emscripton/python) tests are failing?

Because I always forget to run Python and Emscripten tests after I change things. I usually only build and run C++ tests from CLion. They should be fixed now.

PR #270, opened 2024-11-20
[CQT-190] [Infra] Add clang-tidy

rturrado:
.clang-tidy.
- Add an initial version.
- Add `-misc-unused-alias-decls`, as they seem to be false positives.

run_cpp_linters.py:
- Run clang-tidy-18.
- Add BUILD_FOLDER optional parameter.

Fix some clang-tidy issues across the code base.


elenbaasc (commented):
Are the TODOs in the description still to be part of this PR?

docs/dev-guide/dev-guide.md
```diff
@@ -137,12 +137,18 @@ Continuous Integration will fail if the files do not adhere to a series of forma
...
 python3 ./scripts/run_cpp_linters.py .
 ```
 
 !!! note
 
-    The linters require `clang-format-18` and `clang-tidy-18` to be installed on the system.
+    - The linters require`clang-format-18` and `clang-tidy-18`. 
+    - It is mandatory to have a build before running the linters.
+        - `clang-tidy` expects to find a `compile_commands.json` in a build folder.
+    - It is recommended to build with _gcc_ and in _Release_ mode.
```
elenbaasc, line 149:
```suggestion
    - It is recommended to build with _gcc_ in _Release_ mode.
```

scripts/run_cpp_linters.py
```diff
@@ -9,19 +9,20 @@
...
     usage_str = '''
 Usage:
-    python3 run_cpp_linters.py <ROOT_FOLDER>
+    python3 run_cpp_linters.py <ROOT_FOLDER> [<BUILD_FOLDER>]
 Where:
     ROOT_FOLDER: folder containing the .clang-format, .clang-tidy, and C++ files.
+    BUILD_FOLDER: build folder (optional, defaulted to 'build/Release').
 Example:
-    python3 run_cpp_linters.py .
+    python3 run_cpp_linters.py . cmake-build-debug-clang-wsl
```
elenbaasc, line 17: Is the second argument the name of a build folder? Seems like quite an involved name...

scripts/run_cpp_linters.py
```diff
@@ -32,35 +33,46 @@ def get_list_of_cpp_files(root_folder: os.PathLike) -> list[str]:
     return ret
 
 
-def run_clang_format(root_folder: os.PathLike):
+def run_clang_format(root_folder: os.PathLike, file_list_str: str) -> None:
```
elenbaasc, line 36: Generally you wouldn't mention the type in the variable name, e.g.
```suggestion
def run_clang_format(root_folder: os.PathLike, files: str) -> None:
```

Up to you to decide what you think is best here.

scripts/run_cpp_linters.py
```diff
@@ -32,35 +33,46 @@ def get_list_of_cpp_files(root_folder: os.PathLike) -> list[str]:
...
+    file_list = get_list_of_cpp_files(root_folder)
+    file_list_str = " ".join(file_list)
+    run_clang_format(root_folder, file_list_str)
+    run_clang_tidy(root_folder, build_folder, file_list_str)
 
 
 def main(argv):
     logging.basicConfig(level=logging.INFO, format='%(message)s')
-    if len(argv) != 2:
+    if len(argv) < 2 or len(argv) > 3:
```
elenbaasc, line 70: Why doesn't `len(argv) != 2` work?

src/v3x/instruction.cpp
```diff
@@ -23,7 +23,7 @@ Instruction::Instruction(std::string name, const std::optional<std::string>& ope
  * Equality operator.
  */
 bool Instruction::operator==(const Instruction& rhs) const {
-    return name == rhs.name && operand_types == rhs.operand_types;
+    return name == rhs.name && operand_types.equals(rhs.operand_types);
```
elenbaasc, line 26: Why use the equality operator and the function? Does the latter perform a different kind of check?

test/v3x/cpp/integration_test.cpp
```diff
@@ -15,7 +15,6 @@
 
 namespace cqasm::v3x::test {
 
-namespace cq3x = cqasm::v3x;
```
elenbaasc, line 18: Thanks! Would we, at a later stage, want get rid of any mention of cqasm v3 altogether?

elenbaasc (dismissed):
Looks all good!

Just some questions.

rturrado (commented):
Thanks for the comments!

scripts/run_cpp_linters.py
elenbaasc: Up to you 😄

src/v3x/instruction.cpp
elenbaasc: Ah, I see, thanks for 'pointing' it out.

test/v3x/cpp/integration_test.cpp
elenbaasc: I guess something to think about later. I mean ultimately we would have different versions of libQASM being compatible with different versions of cQASM 3.x. But this doesn't have to be decided now.

rturrado: Yes, the world of software versioning is something that I don't know much about, I'm sure it can be quite interesting. But it's a bridge we are not going to cross anytime soon.

PR #272, opened 2024-12-02
[CQT-217] Error when redeclaring a variable

rturrado:
Up until now, a variable could be redeclared.

Upon a redeclaration, we would simply:
- remove the previous entry in the variable symbol table, and
- add the new variable to the symbol table.

But this was leading us to different kind of problems.

From now on, a program as the one below will emit an error message: `trying to redeclare variable 'q'`.

```
version 3.0

qubit[2] q

H q[0]

qubit[3] q

H q[2]
```


elenbaasc (approved):
Perfect, thanks!

PR #280, opened 2025-02-26
[CQT-347] Add support for assembly declarations

rturrado:
Add `ASM` and `RAW_STRING_TEXT` tokens.
Add `asm-declaration` parser rule.
Add `asm_declaration` syntactic and semantic tree nodes.

### DIFF
--- a/src/v3x/CqasmLexer.g4
+++ b/src/v3x/CqasmLexer.g4
@@ -1,5 +1,9 @@
 lexer grammar CqasmLexer;
 
+@members {
+    int unmatched_inner_open_braces = 0;
+}
+
 // White spaces and comments are skipped, i.e. not passed to the parser
 WHITE_SPACE: [ \t]+ -> skip;
 SINGLE_LINE_COMMENT: '//' ~[\r\n]* -> skip;
@@ -58,6 +62,7 @@ POW: 'pow';
 CTRL: 'ctrl';
 QUBIT_TYPE: 'qubit';  // types
 BIT_TYPE: 'bit';
+ASM: 'asm'  -> pushMode(ASSEMBLY_DECLARATION);  // assembly declaration
 
 // Numeric literals
 BOOLEAN_LITERAL: 'true' | 'false';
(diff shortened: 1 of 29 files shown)


src/v3x/semantic_analyzer.cpp
```diff
@@ -296,6 +296,26 @@ std::any SemanticAnalyzer::visit_non_gate_instruction(ast::NonGateInstruction& n
...
 
+std::any SemanticAnalyzer::visit_asm_declaration(ast::AsmDeclaration& node) {
+    auto ret = tree::make<semantic::AsmDeclaration>();
+    try {
+        ret->backend_name = node.backend_name->name;
+        ret->backend_code = node.backend_code;
+
+        // Copy annotation data
+        ret->annotations = std::any_cast<tree::Any<semantic::AnnotationData>>(visit_annotated(*node.as_annotated()));
+        ret->copy_annotation<parser::SourceLocation>(node);
```
elenbaasc, line 307: Just a question: Does the current version of libQASM support annotations?

elenbaasc (commented):
Looks all good, thanks!

Just minor questions.

src/v3x/CqasmLexer.g4
```diff
@@ -73,10 +75,14 @@ fragment Exponent: [eE][-+]?Digit+;
...
+fragment RawText: (.)*?;
+
 // Version mode
 //
 // Whenever we encounter a 'version' token, we enter the Version mode
-// Within the version mode, a sequence such as '3.0' will be treated as a version number, and not as a float literal
+// Within the Version mode, a sequence such as '3.0' will be treated as a version number, and not as a float literal
 mode VERSION_STATEMENT;
-VERSION_WHITESPACE: [ \t]+ -> skip;
+VERSION_WHITE_SPACE: [ \t]+ -> skip;
```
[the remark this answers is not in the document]
rturrado: Haha, much simpler than that. It's because in line 4 I use the term `WHITE_SPACE`.

```
// White spaces and comments are skipped, i.e. not passed to the parser
WHITE_SPACE: [ \t]+ -> skip;
```

src/v3x/semantic_analyzer.cpp
rturrado: Indeed.

This has to do with tree-gen. We use tree-gen to automatically generate code from a `.tree` file. This code is basically a set of classes. For example: a semantic AST node, such as a statement, a gate, and so on. In the `.tree` file we just specify how these classes are related in a hierarchy, and what fields each class has. For example, in `semantic.tree`, we have the node `gate`, which will turn into a C++ `Gate` class:

```
    gate {
        name: cqasm::v3x::primitives::Str;
        gate: Maybe<gate>;
        parameter: external Maybe<cqasm::v3x::values::ValueBase>;
    }
```

The types of each field can be: a primitive value, like `Str`, or another AST node, like `gate`, or a reference to an AST node defined in another `.tree` file, like `ValueBase`. tree-gen also uses its own versions of standard types (containers, smart pointers...). For example, in the example below, `Maybe` is an optional node.

Now, if we look at the hierarchy of nodes in a semantic AST, we will see something like: Base < Node < Annotated < ...
This is, there is a base class `Base`, from which every `Node` in the AST inherits, and then there is a whole bunch of nodes under `Annotated`. For example, a `statement` and a `variable` can be annotated. And that means that those nodes can include annotation data. At the moment, the only annotation data that we are using is the source location (where in the source program, line and column, that node starts and ends). That information is retrieved and used in case of parsing errors.

elenbaasc: Thanks for clearing that up!

elenbaasc (approved):
Looks good!

Thanks for swift implementation and explanation!

elenbaasc (approved):

PR #281, opened 2025-02-27
Do not hardcode CMAKE_CXX_STANDARD

rturrado:
This is a suggestion coming from this [comment](https://github.com/conan-io/conan-center-index/pull/26499#pullrequestreview-2647168875).

> Following best CMake best practices from Professional CMake 19th edition:
> 
> > For a standalone project that will always be built as the top level project, hard-coding variables like
CMAKE_CXX_STANDARD and CMAKE_CXX_EXTENSIONS may be appropriate. But some projects may end up
being absorbed into a larger parent project (see Chapter 39, FetchContent), and that parent project
may want to enforce a higher language standard or make use of extensions that this project doesn’t
need.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - **Removed** for now removed features.
 
 
+## [ 1.0.1 ] - [ xxxx-yy-zz ]
+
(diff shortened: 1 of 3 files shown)


elenbaasc (approved):
Great!

elenbaasc (approved):

PR #283, opened 2025-02-27
Rename `ast` to `syntactic`.

rturrado:
In libQASM, `ast` stands for Abstract Syntax Tree, but logically refers to the syntactic phase of the parser.

So on one hand we have `ast` namespace, and `ast.tree` and `ast.{hpp,cpp}` files, but on the other hand we have `semantic` namespace, and `semantic.tree` and `semantic_analyzer.{hpp,cpp}`.


rturrado:
> Will this require an update on the side of the frontend?

No because the clients only use the semantic AST, not the syntactic AST. So this change is completely internal.

elenbaasc (approved):

elenbaasc (approved):
Looks good to me

PR #284, opened 2025-03-03
Do not use ANTLR4 Conan package

rturrado:
And update ANTLR4 git tag to the latest one, which includes a fix for Windows/debug builds.


rturrado:
We are going to temporarily disable the Windows/debug builds both for libQASM and QX.
This as a workaround to the problems of using antlr4-cppruntime with the latest Visual Studio compiler.

PR #285, opened 2025-03-03
Add initial package.json definition

Siilwyn:
https://qutech-sd.atlassian.net/browse/QI2-1474?atlOrigin=eyJpIjoiZTAyYzZmZWNlYmFhNGRiMjk5MjJmNzk5OGYzNGNmMGUiLCJwIjoiaiJ9

Untested but should be able to publish an npm package with this by running it in the build pipeline that provides the JS & wasm files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,11 +6,11 @@
     "url": "https://github.com/QuTech-Delft/libqasm.git"
   },
   "files": [
-    "build/Release/emscripten/cqasm_emscripten.js",
-    "build/Release/emscripten/cqasm_emscripten.wasm"
+    "./build/Release/emscripten/cqasm_emscripten.js",
+    "./build/Release/emscripten/cqasm_emscripten.wasm"
   ],
   "exports": {
-    "initWrapper": "build/Release/emscripten/cqasm_emscripten.js",
-    "cqasm.wasm": "build/Release/emscripten/cqasm_emscripten.wasm"
+    "initWrapper": "./build/Release/emscripten/cqasm_emscripten.js",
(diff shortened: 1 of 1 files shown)


oschusler:
@rturrado I've asked @Siilwyn to add this. At the moment, the WASM file is copied to the platform for every new version of libQASM. This is not a maintanable solution. With this PR, we can push the libQASM WASM file to npm (the package registry for Node packages). That way, Renovate can automatically pick up on changes in the future.

rturrado:
> @rturrado I've asked @Siilwyn to add this. At the moment, the WASM file is copied to the platform for every new version of libQASM. This is not a maintanable solution. With this PR, we can push the libQASM WASM file to npm (the package registry for Node packages). That way, Renovate can automatically pick up on changes in the future.

Perfect. Thanks for the explanation. I'm perfectly fine as it is.

I was just wondering if you consider it worth adding some type of documentation.

Also, more importantly, I would like to know how to update this file with new libQASM versions. Should one just update the  file directly, pointing to the last libQASM version?

Siilwyn:
Hi @rturrado so to clarify this PR needs some additional work I need help with :)

The CI pipeline should run `npm publish` after it created the needed WASM artifacts, but only when a new version is released. When a new version is released the `version` field in the `package.json` has to be set to the new version number.

Regarding documentation it would be very nice to add a link to the npm page once it's published. E.g: https://www.npmjs.com/package/css-declaration-sorter

rturrado (approved):

PR #286, opened 2025-03-05
Disable Windows/Debug build.

rturrado:
Disable temporarily.
Until an antlr4-cppruntime>4.13.1 Conan package builds in Windows/Debug.


elenbaasc (approved):
Bon!

PR #287, opened 2025-03-05
Fix `build_type` compile flag

rturrado:
Set build_type compile option correctly.
This is the reason why Windows/debug builds were working in libQASM but not on QX simulator.


rturrado:
> But I guess then also for libQASM the Windows/debug builds need to be disabled, momentarily, right?

Yes. The other PR goes first. Then this one.

PR #290, opened 2025-03-10
Fix `cibw-wheels` job

rturrado:
Current `Build wheels` step in `cibw-wheels` job is only being executed for Linux/ARM64, because of the `if` clause.
Remove the `if` clause.
Set `CIBW_BEFORE_ALL_LINUX` to the Java installation command for Linux/ARM64.
Set `CIBW_BEFORE_ALL_LINUX` to a no-op otherwise.
The name of this branch starts with 'release' in order to test the `assets` workflow.


elenbaasc (commented):
Can we test this somehow (without making an actual release)?

rturrado:
> Can we test this somehow (without making an actual release)?

Yes. You have the generated wheels available as workflow artifacts: https://github.com/QuTech-Delft/libqasm/commit/7be4c810fa2f89e0a5b0f961aba8a1e8b0639141/checks?check_suite_id=35470775592

rturrado:
> OK. What we can do is:
> 
> * Rename the current branch and merge it into `develop`.

Actually, this step is a bit of a nonsense. Why would we rename it? Since it doesn't do any harm as it is.

elenbaasc:
> Actually, this step is a bit of a nonsense. Why would we rename it? Since it doesn't do any harm as it is.

If there are no unwanted side-effects (e.g., that a publishing pipeline is started based of the name of the branch) then that should be fine. 

The issue now is that the binaries of version 1.0.0 are missing (it's not that they are all there and contain a bug), so we should create these (by releasing them from the 1.0.0 tagged commit) and upload them manually to PyPI. We could release a patch, but since we already have large features (e.g. the asm declaration and Rn gate), we might as well release it all together in 1.1.0 once were ready.

rturrado:
> If there are no unwanted side-effects (e.g., that a publishing pipeline is started based of the name of the branch) then that should be fine. 

You don't publish anything unless you create a release in GitHub.

> The issue now is that the binaries of version 1.0.0 are missing (it's not that they are all there and contain a bug), so we should create these (by releasing them from the 1.0.0 tagged commit) and upload them manually to PyPI. We could release a patch, but since we already have large features (e.g. the asm declaration and Rn gate), we might as well release it all together in 1.1.0 once we're ready.

OK. Let's merge this into`develop`. Then port to `master` only the changes in `assets.yaml` here, and create the binaries so that you can manually upload them.

elenbaasc (approved):
Look good, thanks!

PR #291, opened 2025-03-12
Fix `cibw-wheels` job in `master`

rturrado:
Current 'Build wheels' step in cibw-wheels job is only being executed for Linux/ARM64, because of the 'if' clause.
Remove the 'if' clause.
Set CIBW_BEFORE_ALL_LINUX to the Java installation command for Linux/ARM64.
Set CIBW_BEFORE_ALL_LINUX to a no-op otherwise.
The name of this branch starts with 'release' in order to test the assets workflow.


rturrado:
We won't be merging this PR. We just used it to generate all the wheels that should have been created with release 1.0.0.

PR #295, opened 2025-03-26
[CQT-358] Check that qubit operands of a gate instruction have the same number of indices

rturrado:
Add `check_qubit_operands_indices_have_same_size` and call it from `visit_gate_instruction` in the semantic analyzer.
Add `cnot_q_0_q_2_3` integration test.
Update CHANGELOG.md.


elenbaasc (approved):
Looks all good to me, thanks!

PR #298, opened 2025-07-12
[CQT-386] Semantic analyzer does not check if number of parameters is correct

elenbaasc:
- Added parameter size check in `semantic_analyzer.cpp`.
- Added `Python` and `C++` tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - **Fixed** for any bug fixes.
 - **Removed** for now removed features.
 
+## [ M.m.P ] - [ xxxx-y-zz ]
+
+### Added
+- Semantic analyzer checks if the number of provided instruction parameters are correct.
+
 
 ## [ 1.2.0 ] - [ 2025-03-27 ]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [ M.m.P ] - [ xxxx-y-zz ]
 
-### Added
-- Semantic analyzer checks if the number of provided instruction parameters are correct.
+### Fixed
+- Semantic analyzer throws exception if the number of provided instruction parameters are incorrect.
 
 
 ## [ 1.2.0 ] - [ 2025-03-27 ]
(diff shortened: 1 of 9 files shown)


src/v3x/semantic_analyzer.cpp
rares1609, line 1: This code is in accordance to the main objective of the issue, checking whether non-parametrized gates are given parameters, and whether parametrized gates are given more parameters than their constructor expects.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - **Fixed** for any bug fixes.
 - **Removed** for now removed features.
 
+## [ M.m.P ] - [ xxxx-y-zz ]
+
+### Added
+- Semantic analyzer checks if the number of provided instruction parameters are correct.
```
rares1609: I could be wrong, but wouldn't this issue better fall under the `###Fixed` category in the changelog?

rares1609 (commented):
The issue looks good to me. It meets all the requirements in the Jira ticket, namely having the `resolve_parameters` function check the number of parameters in parametrized gates (or no params for non-parametrized gates) and adding the changelog entry.

There are only 2 comments I would have:


1.  The additions made in this issue, I think would be better marked under the `###Fixed` category in `CHANGELOG.md`. Of course, leave it as is if you think otherwise.
2. When it comes to the `ast.golden.txt` files for the `H` and `Rn` gates, doesn't the `Rx` gate need something similar? I am referring to a similar file at `res/v3x/tests/integration/instruction/Rx_param/ast.golden.txt`. Also, same thing for the `input.qc` and `semantic.3.0.golden.txt` files that would go in the `Rx_param` folder.

CHANGELOG.md
elenbaasc: It's a bit of a philosophical question... I'm not sure. If you didn't catch an error before, but the functionality does not change after, have you really fixed something... It wasn't broken, it just was not informing the user that something wasn't right. I guess the intended behavior does change, now that you get an error for things a user might have thought to be valid.

I will put it under fixed... :P

elenbaasc:
I left out the Rx gate from the cpp tests, because it is not very necessary. I'm testing a non-parameterized gate the H and a parameterized gate Rn, whether the parameterized gate has one or more parameters, it doesn't really matter as the exception that is thrown will be exactly the same. Having the Rx gate in the python tests is therefore a bit overkill, but I thought I would just leave it.

rares1609 (approved):

PR #299, opened 2025-07-17
[CQT-392] Release for Python 3.13

elenbaasc:
- Update to support Python 3.13.
- Update `cibuildwheel` package in tests to test include builds for Python 3.13.


rares1609 (approved):
This looks good! `Python 3.13` is added in the relevant files, and an entry is added to the changelog.

PR #300, opened 2025-07-28
Creating release 1.2.1

elenbaasc:
### Added
- Support for Python 3.13.

### Fixed
- Semantic analyzer throws exception if the number of provided instruction parameters are incorrect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - **Fixed** for any bug fixes.
 - **Removed** for now removed features.
 
-## [ M.m.P ] - [ xxxx-y-zz ]
+## [ 1.2.1 ] - [ 2025-07-28 ]
+
+### Added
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -19,7 +22,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - Semantic analyzer check for same number of indices in each qubit operand.
```
rares1609, line 24: I see there are two `### Added` columns. Should the note on the semantic analyzer be added to the first column, and delete the second one?

rares1609 (approved):
All looks good! I only have a comment about the CHANGELOG, but I will still mark it as approved if you think it should be as is now.

CHANGELOG.md
elenbaasc: For this release there is only one `### Added` column. The other column is part of the `1.2.0` release.

rares1609: My bad, did not read properly

PR #311, opened 2026-04-21
[CQT-450] Update libQASM

elenbaasc:
Added the following two-qubit gates:

- `CY`
- `SqrtSWAP`
- `InvSqrtSWAP`
- `ISWAP`
- `SqrtISWAP`
- `MS`
- `DCNOT`
- `CV`
- `M`
- `ECR`

Added the following measure instruction variants (and tests for these variants):

- `measureX`
- `measureY`
- `measureZ`


rares1609 (approved):
This looks good to me. The mentioned two-qubit gates and measure instructions are implemented, with the latter being tested too. 

The one thing which sprang to mind was that maybe tests should be made for the two-qubit gates too? At the same time, I looked a bit and at first glance I did not see tests for other gates, so maybe this is done on purpose. To this end I will leave it approved, as is. 👍

